PR #65, opened 2015-01-12
remove only the indices, not the square brackets

Dakuan:
The `{indices: false}` feature isn't really doing what you would expect from the API. It's removing both the index number and the square brackets. This is a shame as the square brackets are expected to be there by a great many web technologies including PHP and Ruby On Rails.

### DIFF
--- a/lib/stringify.js
+++ b/lib/stringify.js
@@ -41,8 +41,8 @@ internals.stringify = function (obj, prefix, options) {
         var key = objKeys[i];
         if (!options.indices &&
             Array.isArray(obj)) {
-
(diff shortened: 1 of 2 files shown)


nlf:
This would be a breaking change, instead let's add a new option `brackets` that will enable or disable brackets when `indices === false` and default it to `false`, that way the old behavior remains but can be explicitly overridden.

Dakuan:
@nlf had another suggestion: https://github.com/hapijs/qs/issues/46

nlf:
Superseded by #70, closing this

Issue #47, opened 2014-10-06
Array of objects parsing


dmitriyles:
I'm trying to parse an IPN message from PayPal, and it has an array of objects in a query string:
```url
transaction%5B0%5D.is_primary_receiver=true&transaction%5B0%5D.id_for_sender_txn=5GN104358V194442M&log_default_shipping_address_in_transaction=false&transaction%5B0%5D.receiver=seller1.test%40mercher.net&action_type=PAY&ipn_notification_url=http%3A//staging.mercherdev.com/ipn&transaction%5B1%5D.paymentType=SERVICE&transaction%5B0%5D.amount=USD+136.25&charset=windows-1252&transaction_type=Adaptive+Payment+PAY&transaction%5B1%5D.id_for_sender_txn=9L346159YT798301J&transaction%5B1%5D.is_primary_receiver=false&transaction%5B0%5D.status=Completed&notify_version=UNVERSIONED&transaction%5B0%5D.id=3D429284U5773243B&cancel_url=http%3A//staging.mercherdev.com/test&transaction%5B1%5D.status_for_sender_txn=Completed&transaction%5B1%5D.receiver=dmitriy.s.les-facilitator%40gmail.com&verify_sign=AFcWxV21C7fd0v3bYYYRCpSSRl31Aaz5mmivhYQQvF-Zz1csgdIo0LYX&sender_email=buyer1.test%40mercher.net&fees_payer=PRIMARYRECEIVER&transaction%5B0%5D.status_for_sender_txn=Completed&return_url=http%3A//staging.mercherdev.com/test&transaction%5B0%5D.paymentType=GOODS&transaction%5B1%5D.amount=USD+2.50&reverse_all_parallel_payments_on_error=true&tracking_id=32&transaction%5B1%5D.pending_reason=NONE&pay_key=AP-8LF09716FJ765462U&transaction%5B1%5D.id=43253900X5221472M&transaction%5B0%5D.pending_reason=NONE&status=COMPLETED&transaction%5B1%5D.status=Completed&test_ipn=1&payment_request_date=Mon+Oct+06+09%3A16%3A01+PDT+2014
```
`Qs.parse()` gives me this result:
```json
{
    "transaction":                                 [ "NONE", "Completed" ],
    "log_default_shipping_address_in_transaction": "false",
    "action_type":                                 "PAY",
    "ipn_notification_url":                        "http://staging.mercherdev.com/ipn",
    "charset":                                     "windows-1252",
[...]

PR #66, opened 2015-01-13
Add mixed array and object dot notation support Closes: #47

AlexMeah:


lib/parse.js
```diff
@@ -83,6 +83,10 @@ internals.parseKeys = function (key, val, options) {
         return;
     }
 
+    // Transform dot notation to bracket notation
+
+    key = key.replace(/\.(\w+)/g, '[$1]');
```
nlf: This feels too strict, `\w` is alphanumeric plus underscores, so it rules out plenty of valid characters. Maybe better to make it `/\.([^\.\[]+)/`.

I'm also curious how big of a hit this has on performance..

nlf:
I'm trying to think of scenarios this might not work, but this is definitely the simplest solution for 90+% of use cases.

I'm still undecided how flexible I want this module to be, since there's been a *ton* of bikeshedding and debate over what it should do and how. Part of me thinks it should support as many use cases as possible, but the other part of me realizes it borders on crazy to do that and I should just keep what it does clearly defined and move on.

I'm still chewing on this, and I'll get back to it, but thanks for the effort so far!

lib/parse.js
AlexMeah: I have updated the regex.

What is the best way to measure perf? A few console logs show the speed to be the same as before.

AlexMeah:
We bumped into this issue when dealing with a legacy .net application which has to remain unchanged. 

We need to being accepting form submissions using express-body-parser and would like to avoid using a custom implementation on specific routes.

Nijikokun:
:+1:

AlexMeah:
Any movement on this?

hueniverse:
@nlf this is a 4 months old pull request. Please merge or close.

PR #76, opened 2015-03-18
RFC 3986

jamie-beck:
This is a patch to get closer to the RFC 3986 specs as described here:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent

By following the RFC more closely the resulting string will be more cross language (programming) compatible.


nlf:
Nice, thanks

PR #78, opened 2015-03-25
Remove useless entry point.

ben-eb:
This file is not needed to load the module. Just specify a main entry point in `package.json` and browserify/node will resolve the module automatically.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,1 +1,0 @@
-module.exports = require('./lib/');
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "version": "2.4.1",
   "description": "A querystring parser that supports nesting and arrays, with a depth limit",
   "homepage": "https://github.com/hapijs/qs",
-  "main": "index.js",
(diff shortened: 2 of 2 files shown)


hueniverse:
While useless, this is the current style across all hapi.js projects. Removing it really adds no value. This change is as much noise as the issue it is trying to fix.

ben-eb:
> Removing it really adds no value.

If bundling this module with browserify, the extra entry point adds more weight to the build, however small, which is relevant on mobile devices where data is at a much higher cost. I think it is bad style to include needless files like this when it can be done with less lines of code. But, an alternate change might be to rename `lib/index.js` to `index.js`.

I agree that it is completely irrelevant in a server side context.

nlf:
Unless you can prove that the weight it adds is significant enough to care about (which I highly doubt, seeing as it's one line) this is a no. I'd prefer maintaining consistency with the other hapi.js projects over eliminating a single line of code.

ben-eb:
Fair enough, that is your decision to make.

PR #81, opened 2015-04-21
Add a `filter` option

kitcambridge:
This matches `JSON.stringify()`, which accepts a filter function or property list as its second argument. If a function is specified, it is called to obtain the replacement value. Otherwise, if an array is specified, it is used to select properties and array indices for stringification.

This could be accomplished via a deep copy in user code, but I think it's useful to have—especially since `qs.stringify` traverses the object, anyway.

### DIFF
--- a/README.md
+++ b/README.md
@@ -232,6 +232,32 @@ Qs.stringify({ a: 'b', c: 'd' }, { delimiter: ';' });
 // 'a=b;c=d'
 ```
 
+Finally, you can use the `filter` option to restrict which keys will be included in the stringified output.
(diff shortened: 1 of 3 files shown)


nlf:
sorry for the delay, i'll get this reviewed soon.

kitcambridge:
@nlf No rush. Thanks! :star:

nlf:
@kitcambridge mind updating this so it will merge cleanly and adding some docs? then i'll be happy to merge it and get it into the 3.0.0 release

kitcambridge:
@nlf All set. Thanks again!

PR #85, opened 2015-05-19
No equal sign

pierot:
Pull request for functionality described here: https://github.com/hapijs/qs/issues/79#issuecomment-103473308

### DIFF
--- a/lib/parse.js
+++ b/lib/parse.js
@@ -9,7 +9,8 @@ var internals = {
     delimiter: '&',
     depth: 5,
     arrayLimit: 20,
-    parameterLimit: 1000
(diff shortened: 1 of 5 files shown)


scharf:
Looks good, but I am afraid this change will break a lot of existing code. I think it would be best to hide this feature behind an option (like `strictNullHandling`)

tunnckoCore:
> I think it would be best to hide this feature behind an option

:+1:

nlf:
This definitely needs to be behind an option

pierot:
I updated the PR with the option `strictNullHandling`.

Issue #79, opened 2015-03-26
In 'a&b=' treat `a` as boolean flag (true)


scharf:
If no `=` is given for a flag, it should be handled as boolean (`true`).

For example:

```javascript
Qs.parse('a&b=', {allowBoolean: true});
//  {a: true, b: ''}
```

without the `allowBoolean` option, it should not do this conversion (as it does today):

```javascript
Qs.parse('a&b=');
//  {a: '', b: ''}
```

pierot:
I would argue that the other way around, giving a null as value for a key, should generate a querystring value without the `=`.
``` javascript
Qs.stringify({a: null, b: "test"});
// ?a&b=test
```

Empty string could still behave as it is doing right now:
``` javascript
Qs.stringify({a: null, b: "test", c: ""});
// ?a&b=test&c=
```

scharf:
@pierot That makes a lot of sense!

Because the following fails today, the following assertion fails:

```javascript
assert.deepEqual(
    Qs.parse(Qs.stringify({a: null, b: "test", c: ""})),  // a=&b=test&c=
    {a: null, b: "test", c: ""}                           // {a: "", b: "test", c: ""} 
);
```

pierot:
I made a pull request for this: https://github.com/hapijs/qs/pull/85

nlf:
Closed via #85

PR #87, opened 2015-05-21
Added documentation for the new `strictNullHandling` flag

scharf:


nlf:
nice, thanks for writing that up

scharf:
Thank you for the quick merge, I think this completes #79 and #85.

PR #88, opened 2015-05-23
Cleanup repo style

idanwe:
Following hapijs/contrib#31

* Move scripts to package.json and remove Makefile
* Change main file to lib/index.js

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "version": "3.0.0",
   "description": "A querystring parser that supports nesting and arrays, with a depth limit",
   "homepage": "https://github.com/hapijs/qs",
-  "main": "index.js",
+  "main": "lib/index.js",
   "dependencies": {},
   "devDependencies": {
     "browserify": "^10.2.1",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -11,7 +11,8 @@
     "lab": "5.x.x"
   },
   "scripts": {
-    "test": "make test-cov",
+    "test": "node node_modules/lab/bin/lab -a code -t 100 -L",
```
Marsup: NPM already has node_modules binaries in its path, simply call lab.

idanwe: I know it, I follow [hapi package.json](https://github.com/hapijs/hapi/blob/master/package.json#L53-L55)
Do you think that there is reason for the `node_modules` path there?

Marsup: I have no idea. @hueniverse ?

hueniverse: Copied from the Makefile... if it works without take it out (and open a PR on hapi)

Marsup: FWIWI this helps only on windows, because it doesn't understand shebangs, but I don't think we care.

nlf: IIRC npm installs a batch script that shims the bin to work in windows, so just pointing directly to lab should work just fine

Marsup: Yeah right, bad memory from trying to make lab work on windows the other day, I had to do that, my bad :)

PR #96, opened 2015-06-17
Throwing an exception instead of converting data structure

nfroidure:


nlf:
There's no way I'd merge this as is.

Let's put it behind an option instead. I can see where some people may want this behavior, but so far they seem to be the minority. A simple flag that allows throwing an error instead of converting to an object is the best way to handle this, I think.

AdriVanHoudt:
flag would indeed be a better solution

nfroidure:
I disagree but it's ok, not my project ;).

PR #97, opened 2015-06-26
decode diacritics

missinglink:
heya, I was expecting this test case to pass, am I doing something wrong or is it a valid bug?

### DIFF
--- a/test/parse.js
+++ b/test/parse.js
@@ -142,6 +142,7 @@ describe('parse()', function () {
 
         expect(Qs.parse('a[b%20c]=d')).to.deep.equal({ a: { 'b c': 'd' } }, { prototype: false });
         expect(Qs.parse('a[b]=c%20d')).to.deep.equal({ a: { b: 'c d' } }, { prototype: false });
+        expect(Qs.parse('place=urn%E4sch')).to.deep.equal({ place: 'urnäsch' }, { prototype: false });
         done();
     });
 


rlidwka:
I would expect it to fail since `0xE4` is not a valid utf8:

```js
> decodeURIComponent('urn%E4sch')
URIError: URI malformed
> decodeURIComponent('urn%C3%A4sch')
'urnäsch'
```

missinglink:
this is how `angular.$http` decided to encode it

```bash
> unescape('urn%E4sch')
"urnäsch"
> unescape('urn%C3%A4sch')
"urnÃ¤sch"
```

rlidwka:
> this is how `angular.$http` decided to encode it

Well considering that `angular.js` is very well known to break various specifications [\[1\]](https://github.com/fb55/htmlparser2/issues/113), [\[2\]](https://github.com/jadejs/jade/issues/969), etc. - it's not that surprising.

Angular should be an example for everyone really, "look at it to learn which patterns you should avoid in your code".

-----

`escape` / `unescape` encode character by character and so they have a problem with unicode. It is solved by adding a non-standard syntax for it like `%uXXXX`:

```js
> escape('…')
'%u2026'
```

However, it's not supported by neither HTTP nor any other standard out there. This is how it should be encoded:

```js
> encodeURIComponent('…')
'%E2%80%A6'
```

So those two methods are kinda deprecated (put into Annex B) in ES5: https://es5.github.io/#B.2

-----

This is why it's not a valid bug.

It could be considered a feature request to add unescaping for 0x80-0xFF characters, but I'm not sure if it's good idea or not, considering that it's unclear what to do when a string has both utf8-encoded and this stuff like `%C3%A4%C3` - is it 3 characters or 2?.

missinglink:
fair enough, well put

rlidwka:
For the reference, here is how node.js querystring module parses it:

```js
> querystring.decode('xxx=urn%C3%A4sch')
{ xxx: 'urnäsch' }
> querystring.decode('xxx=urn%E4sch')
{ xxx: 'urn�sch' }
```

`qs` doesn't replace `%E4` with replacement character though. Maybe their behavior should match.

nlf:
I agree this isn't a bug. If `qs` doesn't know how to decode a portion of a string, I just pass it through so the user can decode it however they deem appropriate.

PR #98, opened 2015-07-01
make returning plain objects and allowing prototype overwriting properties optional

nlf:
this is a short term solution before 5.0.0 which will change everything. Essentially the default behavior will match the 2.x releases, with the difference that you may set the `plainObjects` option to `true` to get the behavior from the 3.x releases.

If you don't want plain objects but also don't want to lose keys that would overwrite prototype properties, you can set the `prefixPrototypes` option to `true` and those keys will be prefixed with an underscore `'_'` and passed along. It's not perfect, but at least the keys aren't silently ignored.

Would appreciate your feedback on this one @hueniverse @dougwilson

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,11 +41,11 @@ Qs.parse('a.hasOwnProperty=b', { plainObjects: true });
 // { a: { hasOwnProperty: 'b' } }
 ```
 
-By default parameters that would overwrite properties on the object prototype are ignored, if you wish to keep the data from those fields either use `plainObjects` as mentioned above, or set `prefixPrototypes` to `true` which will add an underscore `'_'` to the beginning of the key names to allow them to be returned:
+By default parameters that would overwrite properties on the object prototype are ignored, if you wish to keep the data from those fields either use `plainObjects` as mentioned above, or set `allowPrototypes` to `true` which will allow user input to overwrite those properties. *WARNING* It is generally a bad idea to enable this option as it can cause problems when attempting to use the properties that have been overwritten. Always be careful with this option.
 
 ```javascript
-Qs.parse('a.hasOwnProperty=b', { prefixPrototypes: true });
-// { a: { _hasOwnProperty: 'b' } }
+Qs.parse('a.hasOwnProperty=b', { allowPrototypes: true });
+// { a: { hasOwnProperty: 'b' } }
 ```
 
 URI encoded strings work too:
--- a/README.md
+++ b/README.md
@@ -34,11 +34,18 @@ For example, the string `'foo[bar]=baz'` converts to:
 }
 ```
 
-The parsed value is returned as a plain object, created via `Object.create(null)` and as such you should be aware that prototype methods do not exist on it and a user may set those names to whatever value they like:
+When using the `plainObjects` option the parsed value is returned as a plain object, created via `Object.create(null)` and as such you should be aware that prototype methods will not exist on it and a user may set those names to whatever value they like:
 
 ```javascript
-Qs.parse('a.hasOwnProperty=b');
+Qs.parse('a.hasOwnProperty=b', { plainObjects: true });
 // { a: { hasOwnProperty: 'b' } }
+```
+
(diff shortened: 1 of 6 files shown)


dougwilson:
To me, this seems to make the situation potentially worse, especially for user-confusion, because of the following not being the same:
```
$ node -pe 'JSON.parse('"'"'{"hasOwnProperty":"yes"}'"'"')'
{ hasOwnProperty: 'yes' }

$ node -pe 'require("querystring").parse("hasOwnProperty=yes")'
{ hasOwnProperty: 'yes' }

$ node -pe 'require("qs").parse("hasOwnProperty=yes")'
{ _hasOwnProperty: 'yes' }
```

Can there be an option to use the default prototype _and_ just overwrite properties, even if it's something you have to opt into?

dougwilson:
lol, please ignore my above comment, as it was technically only based off reading the source code. I see there is `prefixPrototypes`, may bad!

nlf:
`prefixPrototypes` is what allows things that would overwrite a prototype property at all, by prefixing them with a string.  if it's set to false then prototypes would be ignored like they were in 2.x. I can see your point about confusion though, so I can get rid of that option and add an `allowPrototypes` instead that will let people shoot themselves in the foot if they so wish.

dougwilson:
So I had made that comment in haste :) Let me really look through it to see what real comments I have :)

hueniverse:
Looks fine to me. Can't make everyone happy and I really don't care about prototype names in payload. This is JS. You are fucked anyway you look at it.

dougwilson:
Ok, so this PR seems fine in general. My only real comment is that I don't think the global environment should affect the functionality of the library:
```
$ node -pe 'require("qs").parse("hide=1")'
{ hide: '1' }

$ node -pe 'Object.defineProperty(Object.prototype,"hide",{value:function(){},writable:true});require("qs").parse("hide=1")'
{}
```

JSON.parse and the built-in query string does not have this issue, though:
```
$ node -pe 'Object.defineProperty(Object.prototype,"hide",{value:function(){},writable:true});JSON.parse('"'"'{"hide":"1"}'"'"')'
{ hide: '1' }

$ node -pe 'Object.defineProperty(Object.prototype,"hide",{value:function(){},writable:true});require("querystring").parse("hide=1")'
{ hide: '1' }
```

nlf:
That would be resolved by exchanging `prefixPrototypes` with `allowPrototypes`, the latter being an option to let you overwrite properties on `Object.prototype`.

nlf:
Ok, changed this around so instead of optionally adding a prefix to properties that would overwrite the object prototype it instead optionally lets you shoot yourself in the foot, just like JSON.parse. It's off by default, but it's there.

AdriVanHoudt:
This is only breaking when using `{ prototype: false }` right?

nlf:
this reverts the breaking change made in 3.x, so it's in itself another breaking change, though odds are if you don't use any of the prototype stuff on parsed results you'll never notice.

AdriVanHoudt:
ok, thanks for the explanation

README.md
```diff
@@ -34,10 +34,17 @@ For example, the string `'foo[bar]=baz'` converts to:
...
-The parsed value is returned as a plain object, created via `Object.create(null)` and as such you should be aware that prototype methods do not exist on it and a user may set those names to whatever value they like:
+When using the `plainObjects` option the parsed value is returned as a plain object, created via `Object.create(null)` and as such you should be aware that prototype methods will not exist on it and a user may set those names to whatever value they like:
 
 ```javascript
-Qs.parse('a.hasOwnProperty=b');
+Qs.parse('a.hasOwnProperty=b', { plainObjects: true });
+// { a: { hasOwnProperty: 'b' } }
+```
+
+By default parameters that would overwrite properties on the object prototype are ignored, if you wish to keep the data from those fields either use `plainObjects` as mentioned above, or set `allowPrototypes` to `true` which will allow user input to overwrite those properties. *WARNING* It is generally a bad idea to enable this option as it can cause problems when attempting to use the properties that have been overwritten. Always be careful with this option.
```
hueniverse: Might be worth to explicitly note this is a security risk.

Issue #91, opened 2015-05-27
Parse booleans & numbers


xpepermint:
```js
// now
Qs.parse('a=true');    // { a: 'true' }
Qs.parse('a=21.31');    // { a: '21.31' }
// but I would expect boolean
var obj = Qs.parse('a=true');    // { a: true }
Qs.parse('a=21.31');    // { a: 21.31 }
```
Currently I use `express-query-boolean` package which parses `true` and `false` strings into boolean but it would be great to have this feature included in `qs`.

nlf:
My current feeling is that this belongs elsewhere. There are so many conversion patterns that could take place that I don't really want to open the can of worms of implementing them all here. Unless there's a large demand for this, I'm going to say no.

Leaving open for discussion, however.

xpepermint:
OK, I made a plugin for now https://github.com/xpepermint/query-types.

davidraleigh:
I think the @xpepermint request is valid. I was disappointed to see that this isn't a feature. That parse assumes all numbers are strings makes it not so helpful.

travist:
This is actually a problem I am facing using the very popular https://github.com/request/request library. What I am running into is that my REST api is outputting booleans as ```true``` and ```false```. However, when a PUT request is made, this library uses this library to convert it to a querystring, which this then turns all of my booleans into strings breaking the app. I am curious why this library is converting booleans and numbers to strings.  Its a pretty easy fix, and I am willing to submit a pull request.

jdiaz5513:
Part of me wants this too, but automatic type coersion here is troublesome. What if you wanted `"true"` as a string? Or `"21.33"` as a string? I often deal with currency amounts formatted as strings to avoid floating point precision loss.

In particular if you're allowing user input to show up in the query string, all someone has to do is type in `true` where you were expecting a string and good luck avoiding a crash there.

At least the current behavior is predictable and very well defined.

travist:
Thanks for the information. I am no longer blocked by this, so I think I can get by without this change. I was actually able to get around this using https://github.com/visionmedia/superagent instead of https://github.com/request/request (which uses querystring for posts). I can only assume that ```superagent``` does not use query string method for posting information, but I have not dug deep enough yet to determine what is actually different. All I know is my JSON object is not altered from a GET to a POST with ```superagent``` vs ```request.js```. This nugget may prove helpful for anyone else running into this same issue with ```request.js```.

DesignByOnyx:
[...]

PR #101, opened 2015-07-07
Fixed an issue where booleans and numbers are converted to strings.

travist:
This fixes the issue described at #91 .


travist:
This may have been a problem on my end....

PR #106, opened 2015-08-04
Add flag `skipNulls` to optionally skip null values in stringify

5outh:
I was getting frustrated having `null` values in a query object and having to replace them all with `undefined` to get the query string to render the way I wanted. This should fix that problem, by adding a `skipNull` flag to `stringify`, which essentially makes null values act as if they were `undefined` for the purposes of stringifying.

### DIFF
--- a/README.md
+++ b/README.md
@@ -315,3 +315,10 @@ Qs.parse('a&b=', { strictNullHandling: true });
 // { a: null, b: '' }
 
 ```
+
+To completely skip rendering keys with `null` values, use the `skipNulls` flag:
+
+```javascript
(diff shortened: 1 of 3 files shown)


nlf:
Interesting. I need to review this still, but I'll take a look

5outh:
Cool, thanks!

5outh:
@nlf Thanks :) This will help me out.

PayteR:
This is great feature, but i need to take it to the next level - i need to ignore empty strings, how to do that? Could you add this feature? thx

ljharb:
@PayteR please file a new issue for a new feature request; this is a 3+ year old PR.

PR #125, opened 2015-11-03
es6. Closes #124

### LINKED ISSUE #124: ES6 style changes and node v4
The hapi world is moving to only support node >= 4.0. This module is part of that world. I want to make sure the express community is aware of it as they are a significant user of this module. Also, this will affect use in the browser in pre-ES6 versions.

@dougwilson

hueniverse:


lib/index.js
```diff
@@ -1,15 +1,15 @@
...
 
-var internals = {};
+const internals = {};
 
 
-module.exports = {
-    stringify: Stringify,
-    parse: Parse
-};
+exports.stringify = Stringify;
```
AdriVanHoudt: @hueniverse just out of curiosity is there a specific reason for doing this?

hueniverse:
More consistent with other hapi modules.

AdriVanHoudt:
Any particular reason they do it? :P
Don't get me wrong I also started with using this style (the object at the bottom) but when the module (or things to export) become large this just makes it more unclear. So I am thinking of moving away from it but was just wondering whether there was a 'real' reason for it other than style/beauty/readability

Issue #139, opened 2015-11-15
Should export ES5-compatible for npm, otherwise babel transform is required externally


sompylasar:
The ES2015-rewritten 6.0.0 has issues when using via webpack. Usually, babel transform is only applied to the app code, and the `node_modules` directory is excluded entirely. This does not work with 6.0.0 compared to ES5-compatible 5.2.0 -- webpack bundles the ES2015 code of the `node_modules/qs` right into the output, which then breaks the uglify step.

Please provide ES5-compatible build without browserify and point `package.json`'s `main` to it.

This can be done in the `prepublish` step. Look [how it's implemented](https://github.com/rackt/redux-router/blob/master/package.json).

Related: #128, #129, #108.

hueniverse:
Just use the previous version for now. They are the same.

sompylasar:
@hueniverse yes, I do this for now.

bdefore:
i ran into this issue while updating my dependencies and saw the latest was 6.0.0 ... unfortunately this caused me to see the error `SyntaxError: let is a reserved identifier` and it was due to the issue @sompylasar describes. this should probably not be published in npm right now.

hueniverse:
I suggest you learn how semver works and don't upgrade to a new major version without understanding what broke.

bdefore:
@hueniverse i am aware of how semver works, thank you. thought you might want to know that there's more than one person that would like to use `qs` latest with other their other ES5 transpiled libraries.

arb:
You can also add the path to qs in your array of included files for the babel transformer. Then you could exclude all of the other node_modules, but include qs.

zerkms:
@arb 

> You can also add the path to qs in your array of included files for the babel transformer

how would one do that actually? If `node_modules` is excluded using `exclude` parameter - you cannot include anything from it.

arb:
When you are setting up the loader, if you are excluding "node_modules", there is also an `include` option where you can pass an array of values. Add the path to QS into that array. (This is just speculating as I haven't done this, but I think that should work.)

```js
module: {
  loaders: [
    {
      test: /\.js?$/,
      exclude: /node_modules/,
      include: [
        'path/to/your/app',
        'node_modules/qs'
      ]
      loader: 'babel',
      query: {
        presets: ['es2015']
      }
    }
  ]
}
```

Something along those lines. You'll have to play around with the include paths but that is the general gist of it.

sompylasar:
[...]

PR #145, opened 2015-12-11
Change package "main" to the dist-version

hylle:
Removes the need for babel (or similar) when require'ing qs for use in the browser.


Marsup:
Not gonna happen. #139

ljharb:
A build process is required for modern JS development. The next major release will drop `bower` and `component` support, as well as removing the "dist" build entirely from the npm package.

PR #146, opened 2015-12-28
Use dist qs instead of uncompiled es6 version

jeremistadler:


jeremistadler:
Don't think build fail is because of the changes

ljharb:
This isn't necessary anymore, as I've removed all of the ES6 stuff from master in 224c8af8fc8abf48fd9ded33a38997a1d942c1c1

The failure is because nothing is running `npm run dist` anymore as part of tests, nor on prepublish. I'll fix that before the next release.

Issue #149, opened 2016-02-01
Parsed array not being compacted on deeper levels


tdzienniak:
This code ilustrates the issue:
```javascript
let qs = require('qs')

let raw = {
    'a[1][b][1][c][1][d]': 'e'
}

let parsed = qs.parse(raw, {
    depth: 6
})
```
Here is stringified result:
```
{
    "a": [
        {
            "b": [
                null,
                {
                    "c": [
                        null,
                        {
                            "d": "e"
                        }
                    ]
                }
            ]
        }
    ]
}
```
Runnable code [here](https://tonicdev.com/tdzienniak/qs-sparse-array-bug).

tdzienniak:
I've made a PR with fix.

PR #150, opened 2016-02-02
Compacting nested sparse arrays

tdzienniak:
Fix for #149

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -132,7 +132,7 @@ exports.compact = function (obj, references) {
         var compacted = [];
 
         for (var i = 0; i < obj.length; ++i) {
-            if (Array.isArray(obj[i]) || Object.prototype.toString.call(obj[i]) === '[object Object]') {
+            if (obj[i] && typeof obj[i] === 'object') {
                 compacted.push(exports.compact(obj[i], refs));
             } else if (typeof obj[i] !== 'undefined') {
                 compacted.push(obj[i]);
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -132,7 +132,9 @@ exports.compact = function (obj, references) {
         var compacted = [];
 
         for (var i = 0; i < obj.length; ++i) {
-            if (typeof obj[i] !== 'undefined') {
+            if (Object.prototype.toString.call(obj[i]) === '[object Object]') {
+                compacted.push(exports.compact(obj[i], refs));
+            } else if (typeof obj[i] !== 'undefined') {
                 compacted.push(obj[i]);
             }
         }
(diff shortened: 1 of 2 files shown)


lib/utils.js
```diff
@@ -132,7 +132,9 @@ exports.compact = function (obj, references) {
         var compacted = [];
 
         for (var i = 0; i < obj.length; ++i) {
-            if (typeof obj[i] !== 'undefined') {
+            if (Object.prototype.toString.call(obj[i]) === '[object Object]') {
```
ljharb: this test won't give true for arrays or functions - perhaps this should test for any non-primitive?

tdzienniak: Check for arrays is fine but check for functions is in my opinion not nescessary, because you can't define functions in query string.

ljharb: ha, that's a great point, i was thinking in generic JS terms.

lib/utils.js
```diff
@@ -132,7 +132,9 @@ exports.compact = function (obj, references) {
         var compacted = [];
 
         for (var i = 0; i < obj.length; ++i) {
-            if (typeof obj[i] !== 'undefined') {
+            if (Array.isArray(obj[i]) || Object.prototype.toString.call(obj[i]) === '[object Object]') {
```
ljharb: i wonder tho if this might still be a simpler check to do `obj[i] && typeof obj[i] === 'object'` which would catch all non-function primitives?

test/parse.js
```diff
@@ -181,6 +181,8 @@ test('parse()', function (t) {
 
     t.test('compacts sparse arrays', function (st) {
         st.deepEqual(qs.parse('a[10]=1&a[2]=2'), { a: ['2', '1'] });
+        st.deepEqual(qs.parse('a[1][b][2][c]=1'), { a: [{ b: [{ c: '1' }] }] });
+        st.deepEqual(qs.parse('a[1][2][3][c]=1'), { a: [[[{ c: '1' }]]] });
```
ljharb: could we also include a test case for `'a[1][2][3][c][1]=1'`?

tdzienniak: Sure, I will add it.

lib/utils.js
tdzienniak: There is a problem with `null` which type is also an `object`. We could add a check for `null` though.

ljharb: that's why the first truthiness check is there :-)

tdzienniak: Yeah, you are right. It's morning... :-)

ljharb:
Thanks!

PR #151, opened 2016-02-03
allowDots option for stringify. Tests also updated.

snow01:
Added allowDots option for stringify. I have added a new test case for the same.

### DIFF
--- a/lib/stringify.js
+++ b/lib/stringify.js
@@ -20,7 +20,7 @@ var internals = {
     encode: true
 };
 
-internals.stringify = function (object, prefix, generateArrayPrefix, strictNullHandling, skipNulls, encode, filter, sort) {
+internals.stringify = function (object, prefix, generateArrayPrefix, strictNullHandling, skipNulls, encode, filter, sort, allowDots) {
     var obj = object;
     if (typeof filter === 'function') {
         obj = filter(prefix, obj);
@@ -65,9 +65,9 @@ internals.stringify = function (object, prefix, generateArrayPrefix, strictNullH
         }
 
         if (Array.isArray(obj)) {
-            values = values.concat(internals.stringify(obj[key], generateArrayPrefix(prefix, key), generateArrayPrefix, strictNullHandling, skipNulls, encode, filter));
+            values = values.concat(internals.stringify(obj[key], generateArrayPrefix(prefix, key), generateArrayPrefix, strictNullHandling, skipNulls, encode, filter, sort, allowDots));
         } else {
-            values = values.concat(internals.stringify(obj[key], prefix + '[' + key + ']', generateArrayPrefix, strictNullHandling, skipNulls, encode, filter));
+            values = values.concat(internals.stringify(obj[key], prefix + (allowDots ? '.' + key : '[' + key + ']'), generateArrayPrefix, strictNullHandling, skipNulls, encode, filter, sort, allowDots));
         }
     }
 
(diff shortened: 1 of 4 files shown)


tdzienniak:
I think that some array stringification test would be good.

snow01:
@tdzienniak -- Following three tests are enough and expected ?

- `{ a: { b: ['c', 'd'] } }  ===> a.b[0]=c&a.b[1]=d`
- `{ a: [{ b: 'c' }] }  ===> a[0].b=c`
- `{ a: [{ b: { c: [1] } }] }  ===> a[0].b.c[0]=1`

lib/stringify.js
```diff
@@ -65,9 +65,9 @@ internals.stringify = function (object, prefix, generateArrayPrefix, strictNullH
         }
 
         if (Array.isArray(obj)) {
-            values = values.concat(internals.stringify(obj[key], generateArrayPrefix(prefix, key), generateArrayPrefix, strictNullHandling, skipNulls, encode, filter));
+            values = values.concat(internals.stringify(obj[key], generateArrayPrefix(prefix, key), generateArrayPrefix, strictNullHandling, skipNulls, encode, filter, sort, allowDots));
```
ljharb: it looks like you caught a minor bug/omission with the missing "sort" param here and below - i'm not sure if that was intentionally omitted or not originally. could you also add tests that cover the presence/absence of this parameter (here and below)?

Alternatively, if you could prepare a separate PR that adds the sort param and the tests for it, i'd prefer to merge that first and keep this PR focused on "allowDots" :-)

snow01:
@ljharb - Yep, I saw 'sort' param missing as a bug due to omission. Existing tests are not capturing that bug. At depth 3 or more, without that sort does not work.

input --
`{ a: 'a', z: { zj: {zjb: 'zjb', zja: 'zja'}, zi: {zib: 'zib', zia: 'zia'} }, b: 'b' }`

Without 'sort' param fix --

`a=a&b=b&z[zi][zib]=zib&z[zi][zia]=zia&z[zj][zjb]=zjb&z[zj][zja]=zja`

With sort param fix --

`a=a&b=b&z[zi][zia]=zia&z[zi][zib]=zib&z[zj][zja]=zja&z[zj][zjb]=zjb`

Committing the test for this case in same PR, hope okay.

ljharb:
Thanks, that's helpful. I'd still prefer a separate PR but I won't make it a sticking point :-)

snow01:
Lint is giving error for 3 reasons -- 

- more number of parameters ===> we may need to pass parameters as opts.
- more number of statements ===> can refactor some.


/github/qs/lib/stringify.js
   23:23  error    This function has too many parameters (9). Maximum allowed is 8    max-params
   77:18  error    Function 'anonymous' has a complexity of 19                        complexity
   77:18  error    This function has too many statements (33). Maximum allowed is 32  max-statements



From PR perspective how do you suggest we do this so lint passes ?

ljharb:
I'm fine just bumping all those limits in `.eslintrc`. I'll worry about tightening them down later.

snow01:
Rather than bumping up values, should I change those settings to warning ?

ljharb:
Nah, warnings are pretty useless imo since they're easy to ignore.

snow01:
Cool. Done. 

Please pass it soon, if possible, so I can move from dependency on my branch to NPM package.

Thanks !

ljharb:
@snow01 Released as v6.1.0

snow01:
Thanks Jordan !

TheSharpieOne:
Needs documentation added to the readme. Had to search the repo to find this to figure out this feature already existed and then view the changes to see how to use it.

ljharb:
@TheSharpieOne a PR or an issue for updating the docs would be appreciated

PR #154, opened 2016-02-26
[Docs] Add usage instructions for browsers in the README

kabirbaidhya:
Add usage instructions for browsers as the README only has usage for node environment and it is difficult to figure out how to use it in browsers (without using something like browserify/webpack). 

I myself had to go to https://github.com/ljharb/qs/blob/master/dist/qs.js and figure out myself how to do that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,6 +10,7 @@ The **qs** module was originally created and maintained by [TJ Holowaychuk](http
 
 ## Usage
 
+For Node: 
 ```javascript
 var qs = require('qs');
 var assert = require('assert');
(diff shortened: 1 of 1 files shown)


ljharb:
It's 2016 - you need to be using something like browserify or webpack for all of your JS dependencies. In the next version, I'll be removing `dist/qs.js` entirely from the repo - so I'd prefer not to document a deprecated practice that will soon be removed.

kabirbaidhya:
I understand your point. I'd also rather prefer using browserify/webpack, and we're doing that for all new projects. But what about some old codebases & legacy applications where it's not possible to migrate the existing code into these and are still relying on `<script>` tags. Don't you want to support them?

ljharb:
I do support them by ensuring that `qs` works in browserify - so those users can run it through `browserify --standalone` and put the output in their legacy `vendor` directory.

That's a process that everyone maintaining legacy apps needs to learn, and I don't think this random project's readme is the right place to educate them about that.

kabirbaidhya:
Okay, I got it :+1:

ljharb:
Thanks for the contribution, though, and I hope you're not discouraged from making future ones! <3

PR #156, opened 2016-03-11
Reduce npm payload size by cleaning up .npmignore file

jwngr:
Thanks for creating a module I used every day! I'm trying to give back in whatever little way I can.

The npm payload delivered from `npm install qs` contains a bunch of files which are not needed for the module to actually work. I cleaned up the [`.npmignore` file](https://docs.npmjs.com/misc/developers#the-packagejson-file) to get rid of stuff which is excluded by default and include the stuff we want. Here is the data for your package before and after this change:

| Current | Updated | Savings |
| :---: | :---: | :---: |
| 132 KB | 80 KB | **39.39%** |

*I got these numbers using `du -sh <directory>`*

[According to `qs`'s npmjs.com page](https://www.npmjs.com/package/qs), `qs` has been downloaded 1,005,566 times in the last day. At 132 KB per download, that is 132.7 GB per day for this module alone! With this change, that number drops by 39.39% to 80.4 GB!

In case you want to double check the resulting package, you can do so locally:

```
# Create a tarball of the resulting package
$ npm pack

# See the file list of that tarball
$ tar -tf qs-<version>.tgz
package/package.json
package/README.md
package/LICENSE
package/CHANGELOG.md
package/CONTRIBUTING.md
package/dist/qs.js
package/lib/index.js
package/lib/parse.js
package/lib/stringify.js
package/lib/utils.js

# Install the tarball as a local node module
$ npm install qs-<version>.tgz
```


jwngr:
I also was curious if both the `lib/` and `dist/` directories are required? It looks like your [`package.json`'s `"main"` points to `lib/index.js`](https://github.com/ljharb/qs/blob/0f8650b324ddecb5a808f5e27a8fe7fc2638c86f/package.json#L10) but could it just as easily point to `dist/qs.js`? If so, we could exclude the `lib/` directory entirely and save even more space.

ljharb:
Thanks for the contribution! Here's my thoughts:

 - Tests must be included. `npm explore foo && npm install && npm test` should always work, for every module.
 - `dist` is a legacy build output file that will be going away entirely with the next version - consumers should be browserifying themselves, not using a copy-paste script file that creates a global var. Not including it in the npm package is just as much of a breaking change as deleting it entirely, so that must be included until the next major version bump.
 - Numbers that sum up bandwith are a problem for npm, not a problem for individual devs or module authors, so that's just not convincing to me.
 - 132K is super tiny and negligible. The metadata npm loads *before* installing your packages is probably a good chunk of that already :-p

jwngr:
Thanks for the quick response! Here are my responses:

> Tests must be included. `npm explore foo && npm install && npm test` should always work, for every module.

It's interesting you say that. I would say it is generally considered a best practice to excludes tests and docs from the npm payload itself. Clearly you want `npm <blah>` to work when you `git clone` a module's repo, but have you ever actually read docs or ran tests for a module after `npm install`ing it? I don't really see a use case for that.

> `dist` is a legacy build output file that will be going away entirely with the next version - consumers should be browserifying themselves, not using a copy-paste script file that creates a global var. Not including it in the npm package is just as much of a breaking change as deleting it entirely, so that must be included until the next major version bump.

Cool, that makes sense to wait until the next major version bump to remove it.

> Numbers that sum up bandwith are a problem for npm, not a problem for individual devs or module authors, so that's just not convincing to me.

That is most definitely true and I included those numbers more for the wow factor than anything :) But npm is no different than any other service and certainly has limited resources and capacity issues. An `npm install` is far from fast nowadays and I think a big portion of that is that we have bloated packages weighing things downs. npm provides some easy tools to alleviate the problem and I think its low effort for us module authors to use them.

> 132K is super tiny and negligible. The metadata npm loads before installing your packages is probably a good chunk of that already :-p

I agree 132 K is tiny and I don't know the specifics of the metadata you reference (do you have any references on that?), but I think that thinking actually proliferates the problem. The way the npm ecosystem is set up with small packages for every task, doing an `npm install request` ends up downloading 72 deep dependencies for a total of `5M`. If every author takes your view, that number won't get any smaller. But if each author is also able to remove ~40% of their payload, that `5M` drops down to `3M`, which is a notable drop. The more important number here in my opinion is the percentage, not the sheer size differential.

Those are just my thoughts on the topic and hopefully explain why I sent you this PR. I'm happy to make any changes you'd like to it and I won't be offended if you close it (although I of course hope you don't)!

ljharb:
I will be fine merging this if you remove `test`, and the eslint config files, from the `npmignore`.

jwngr:
Okay, done. Savings are a paltry 8K (6.06%) with those changes. I'm definitely interested to hear your thoughts on why you want to include those files if you have the time and energy to explain them.

ljharb:
@jwngr I often want to run tests on dependencies to ensure that they pass on my exact system - there's no guarantee that the precise node + npm + OS versions are being tested by the module authors, and being able to run these tests without tracking down github repos is invaluable. Certainly not every module allows for this, but maximizing the number that do helps keep my debugging time to a minimum.

Disk space and bandwith are much more free and infinite than human time (and I can do plenty of other things while waiting for `npm install` so that's not much of a counter :-p)

PR #160, opened 2016-05-06
Feature/encoder decoder

martinheidegger:
This PR adds a `decoder` and `encoder` option for querystrings. These option are important when dealing with servers that process data in a form other than `utf8`. Namely when trying to send a request to Servers that use `shift_jis` as method this is an important feature.

Note: I choose to add an option instead of re-using the `encode` option to only require semver-minor. Also I bumped the limit for the linting because I couldn't figure out how to push this change more elegantly.

### DIFF
--- a/lib/parse.js
+++ b/lib/parse.js
@@ -10,7 +10,8 @@ var defaults = {
     strictNullHandling: false,
     plainObjects: false,
     allowPrototypes: false,
-    allowDots: false
+    allowDots: false,
+    decoder: Utils.decode
 };
 
 var parseValues = function parseValues(str, options) {
@@ -22,14 +23,14 @@ var parseValues = function parseValues(str, options) {
         var pos = part.indexOf(']=') === -1 ? part.indexOf('=') : part.indexOf(']=') + 1;
 
         if (pos === -1) {
-            obj[Utils.decode(part)] = '';
+            obj[options.decoder(part)] = '';
 
             if (options.strictNullHandling) {
-                obj[Utils.decode(part)] = null;
+                obj[options.decoder(part)] = null;
             }
         } else {
-            var key = Utils.decode(part.slice(0, pos));
-            var val = Utils.decode(part.slice(pos + 1));
+            var key = options.decoder(part.slice(0, pos));
+            var val = options.decoder(part.slice(pos + 1));
 
             if (Object.prototype.hasOwnProperty.call(obj, key)) {
                 obj[key] = [].concat(obj[key]).concat(val);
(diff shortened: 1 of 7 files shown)


.travis.yml
```diff
@@ -1,4 +1,6 @@
 language: node_js
+env:
+  - CXX=g++-4.8
```
ljharb: is this just needed for the `iconv` dev dep?

package.json
```diff
@@ -31,6 +31,7 @@
     "eslint": "^2.9.0",
     "@ljharb/eslint-config": "^4.0.0",
     "parallelshell": "^2.0.0",
+    "iconv": "^2.2.0",
```
ljharb: is this a binary dep? i'd prefer to avoid using binary modules, even in dev deps. is there another way to test this?

lib/stringify.js
```diff
@@ -82,6 +83,7 @@ module.exports = function (object, opts) {
     var strictNullHandling = typeof options.strictNullHandling === 'boolean' ? options.strictNullHandling : defaults.strictNullHandling;
     var skipNulls = typeof options.skipNulls === 'boolean' ? options.skipNulls : defaults.skipNulls;
     var encode = typeof options.encode === 'boolean' ? options.encode : defaults.encode;
+    var encoder = encode ? (typeof options.encoder === 'function' ? options.encoder : defaults.encoder) : null;
```
ljharb: it might be a good idea to throw an explicit exception here if `options.encoder` is provided (not undefined), and not a function.

lib/parse.js
```diff
@@ -134,6 +135,7 @@ module.exports = function (str, opts) {
     options.depth = typeof options.depth === 'number' ? options.depth : defaults.depth;
     options.arrayLimit = typeof options.arrayLimit === 'number' ? options.arrayLimit : defaults.arrayLimit;
     options.parseArrays = options.parseArrays !== false;
+    options.decoder = typeof options.decoder === 'function' ? options.decoder : defaults.decoder;
```
ljharb: it might be a good idea to throw an explicit exception here if `options.decoder` is provided (not undefined), and not a function.

test/parse.js
```diff
@@ -325,8 +327,11 @@ test('parse()', function (t) {
 
     t.test('does not blow up when Buffer global is missing', function (st) {
         var tempBuffer = global.Buffer;
+        console.log('A')
```
ljharb: these console logs should probably be removed?

ljharb:
This looks great! (pending comments) bumping the linter complexity requirements are fine for this.

.travis.yml
martinheidegger: Yes.

package.json
martinheidegger: Yes, there is easily another way. I wanted to provide an example that actually shows how to implement a ShiftJis converter (might be a reoccurring issue)

test/parse.js
martinheidegger: Will fix.

package.json
ljharb: I totally understanding adding the exact test for your use case :-) and i'm all for that. Surely there's an easier way tho?

martinheidegger:
Re-pushed squashed commits that use `iconv-lite` instead of `iconv` for tests (non-binary) and added error messages in case `decoder` or `encoder` is a non-function value.

test/parse.js
```diff
@@ -393,4 +394,34 @@ test('parse()', function (t) {
...
+                    last = parts.index + parts[0].length;
+                }
+                return iconv.decode(new Buffer(result), 'shift_jis').toString();
+            }
+        }), { 県: '大阪府' });
+        st.end();
+    });
+
+    t.test('throws error with wrong decoder', function (st) {
+        try {
```
ljharb: instead of try/catch, you can use `t.throws(function () { … }, new TypeError('Decoder has to be a function.'))` here

test/stringify.js
```diff
@@ -261,4 +262,33 @@ test('stringify()', function (t) {
...
+                    result.push(buf.readUInt8(i).toString(16));
+                }
+                return '%' + result.join('%');
+            }
+        }), '%8c%a7=%91%e5%8d%e3%95%7b&=');
+        st.end();
+    });
+
+    t.test('throws error with wrong encoder', function (st) {
+        try {
```
ljharb: same here wrt `t.throws`

lib/stringify.js
```diff
@@ -82,10 +83,16 @@ module.exports = function (object, opts) {
...
     var sort = typeof options.sort === 'function' ? options.sort : null;
     var allowDots = typeof options.allowDots === 'undefined' ? false : options.allowDots;
     var objKeys;
     var filter;
+
+    if (options.encoder !== null && options.encoder !== undefined && typeof options.encoder !== 'function') {
+        throw new Error('Encoder has to be a function.');
```
ljharb: let's make these `TypeError`s instead of just `Error`s

martinheidegger:
Pushed squashed commits that changed Error → TypeError and used st.throws instead of an try catch.

Issue #94, opened 2015-06-05
Inadequate encoding of Buffers


seishun:
Let's say I want to send some percent-encoded binary data.

```js
> Qs.stringify({a: new Buffer([0x10, 0x20])})
'a=%10%20'
```

So far so good.

```js
> Qs.stringify({a: new Buffer([0xff, 0xff])})
'a=%EF%BF%BD%EF%BF%BD'
```

What? I expected `'%FF%FF'`!

The problem is that it simply [calls `toString` on the Buffer](https://github.com/hapijs/qs/blob/0c6f2a6318c94f6226d3cf7fe36094e9685042b6/lib/stringify.js#L34), which attempts to decode it as UTF-8. It doesn't make sense, since it is then [encoded](https://github.com/hapijs/qs/blob/0c6f2a6318c94f6226d3cf7fe36094e9685042b6/lib/utils.js#L84) back into UTF-8. This round-trip could only be lossless if the original Buffer is a UTF-8 encoded string in the first place, in which case one would just use a string instead.

nlf:
I'm not sure I totally understand what end result you're trying to achieve. This module is for parsing HTTP querystring parameters, are you sending binary data in an HTTP request?

seishun:
> are you sending binary data in an HTTP request?

Yes.

nlf:
as a query parameter? that seems like bad design to me. i can look into making buffer decoding configurable, but i'm not entirely sure it'll work like you want it to anyway. `0xff` would still be interpreted as a character, and would be URI encoded as such to be sure that it can be transmitted as a query parameter. Even stringifying the buffer as binary still yields a different result:

```
> encodeURIComponent(new Buffer([0xff]).toString('binary'))
'%C3%BF'
```

seishun:
>as a query parameter?

Yes, in the POST body.

>0xff would still be interpreted as a character, and would be URI encoded as such to be sure that it can be transmitted as a query parameter.

Query parameters are transmitted as percent-encoded byte sequences, so strings have to be encoded in UTF-8. There is no need to do any encoding for Buffers, since they are already byte sequences - they just have to be percent-encoded.

>Even stringifying the buffer as binary still yields a different result:

Sure, you're giving `encodeURIComponent` a string consisting of a single character with value 255, which is "ÿ". Since the character "ÿ" is not ASCII, it becomes multi-byte when encoded in UTF-8.

nlf:
[...]

PR #161, opened 2016-05-06
Feature/custom buffer encoder

martinheidegger:
This PR restructures the code on top of #160 a little in order to make sure that it is possible to implement a custom encoder for Buffers as mentioned in #94

### DIFF
--- a/test/stringify.js
+++ b/test/stringify.js
@@ -288,4 +288,18 @@ test('stringify()', function (t) {
         }, new TypeError('Encoder has to be a function.'));
         st.end();
     });
+
+    t.test('can use custom encoder for a buffer object', {
+        skip: typeof Buffer === 'undefined'
+    }, function (st) {
+        st.equal(qs.stringify({ a: new Buffer([1]) }, {
+            encoder: function (buffer) {
+                if (typeof buffer === 'string') {
+                    return buffer;
+                }
+                return String.fromCharCode(buffer.readUInt8(0) + 97);
+            }
+        }), 'a=b');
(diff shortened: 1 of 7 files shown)


test/stringify.js
```diff
@@ -291,4 +291,16 @@ test('stringify()', function (t) {
             return;
         }
     });
+
+    t.test('can use custom encoder for a buffer object', function (st) {
```
ljharb: let's make sure to skip this test if `Buffer` isn't available, since these tests need to pass in browsers and old node versions too. eg `t.test('blah', { skip: condition }, function (st) {`

ljharb: let's make sure to skip this test if `Buffer` isn't available, since these tests need to pass in browsers and old node versions too. eg `t.test('blah', { skip: condition }, function (st) {`

test/stringify.js
```diff
@@ -261,4 +262,45 @@ test('stringify()', function (t) {
...
+                encoder: 'string'
+            });
+        } catch (e) {
+            st.equal(e.message, 'Encoder has to be a function.');
+            st.end();
+            return;
+        }
+    });
+
+    t.test('can use custom encoder for a buffer object', function (st) {
```
ljharb: let's make sure to skip this test if `Buffer` isn't available, since these tests need to pass in browsers and old node versions too. eg `t.test('blah', { skip: condition }, function (st) {`

martinheidegger:
Pushed rebased version with a skip check for `Buffer`

PR #162, opened 2016-05-09
Added documentation on how to use qs-iconv with qs

martinheidegger:
I recently publish [qs-iconv](https://github.com/martinheidegger/qs-iconv) that should make it very easy to transform querystrings to a different character encoding. This PR adds the documentation on how to use it to `qs` as well since this is likely a reoccurring issue.


ljharb:
Thanks! Could you rebase these down to a single commit?

martinheidegger:
Done.

PR #165, opened 2016-05-17
link build badge to master branch

a0viedo:


ljharb:
How are these different? Doesn't it default to the master branch?

ljharb:
I think it's better not to be explicit - this way, if i change the default branch on github, the badge will follow it, instead of being pegged to master.

PR #166, opened 2016-05-17
[do not merge] make build fail

a0viedo:
oh boy you really have a **lot** of node versions in your travis conf huh...


ljharb:
lol if you're trying to say that a failed non-master build will fail the overall build badge, i think an english response would have been sufficient :-p

a0viedo:
@ljharb I thought it was quick to reproduce, but badge still remains ok. I guess they changed the default behavior:


Still better to be explicit?

ljharb:
(#165)

PR #169, opened 2016-07-28
add a link to cdnjs in the README file

natesilva:
Now that this library is available from cdnjs for client-side/browser use, document it in the README file.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,6 +21,8 @@ var str = qs.stringify(obj);
 assert.equal(str, 'a=c');
 ```
 
+For client-side (browser) use, the library can be loaded from a CDN. See [cdnjs](https://cdnjs.com/libraries/qs) for the URL to include in your `<script>` tag.
+
 ### Parsing Objects
 
 [](#preventEval)


ljharb:
The preferred mechanism is to use browserify/webpack and include it in your normal build process.

I'm fine adding a link to CDNJS, but making it easier for someone to not use a proper build system is doing no favors.

Also, in v7, I'm going to be removing the prebuilt browser build - how will cdnjs work when that happens? If it's going to break, or require maintainer action, then I'd prefer not to add it to the readme.

natesilva:
Hmm, this sounds like a bad idea then.

One of my company's projects is ancient legacy PHP, no build system, so we use the pre-built `qs.js`. That's why I asked cdnjs to host it. Hopefully most people will be on something more modern…

ljharb:
@natesilva https://wzrd.in/standalone/qs@latest might be more useful, and should continue to work moving forward?

natesilva:
That's interesting. I'll check it out. It might be worth putting that in the README, although your concern about making it easier for people to skip a real build system stands.

PR #171, opened 2016-08-07
Second argument to encoder callback

elhay-av:
With Second argument to encoder callback that indicate witch part of object send to encoder (key or value),
it possible to perform a different action depend on current part.

for example i someone want to decode only the right part (url params value, only what in the right side of `=` sign)

### DIFF
--- a/lib/stringify.js
+++ b/lib/stringify.js
@@ -1,6 +1,10 @@
 'use strict';
 
 var utils = require('./utils');
+var encodeParts = {
+    key: 'key',
+    val: 'val'
+};
 
 var arrayPrefixGenerators = {
     brackets: function brackets(prefix) {
@@ -35,7 +39,7 @@ var stringify = function stringify(object, prefix, generateArrayPrefix, strictNu
         obj = serializeDate(obj);
     } else if (obj === null) {
         if (strictNullHandling) {
-            return encoder ? encoder(prefix) : prefix;
+            return encoder ? encoder(prefix, encodeParts.key) : prefix;
         }
 
         obj = '';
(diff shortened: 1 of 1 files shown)


ljharb:
When in a URL would you want to encode the key and value differently?

elhay-av:
The key in most cases are plane string that not need encoding but can contain chars like []-_

So in this case i want to encode only url params values, that may contain any chars.


[http://www.perfectomobile.com/sites/all/themes/perfecto/img/perfecto_email_logo.jpg]<http://www.perfectomobile.com/>

Elhay Avichezer
SW Developer
Phone:
Email: elhaya@perfectomobile.com<mailto:elhaya@perfectomobile.com>


________________________________

ljharb:
What's the target format you're using that allows unencoded characters in one position, but not in another? In URL query strings, any characters that must be encoded in keys must be encoded in values, and vice versa.

elhay-av:
You right.
But i use []_ in parameters key. This characters not has to be encoded.


[http://www.perfectomobile.com/sites/all/themes/perfecto/img/perfecto_email_logo.jpg]<http://www.perfectomobile.com/>

Elhay Avichezer
SW Developer
Phone:
Email: elhaya@perfectomobile.com<mailto:elhaya@perfectomobile.com>


________________________________

ljharb:
Are you saying that in your use case, they *must not* be encoded? They certainly *can* be encoded in typical systems, and `qs` has always encoded them.

elhay-av:
take this url for example 
all keys `jobNumber[0]`, `startExecutionTime[0]`... not needed to be encoded.
and i want to see this short and nice url:
```
?startExecutionTime[0]=1462690863552&startExecutionTime[1]=custom&status[0]=PASSED&status[1]=FAILED
```

instead of this long and not clear url:
```
?startExecutionTime%5B0%5D=1462690863552&startExecutionTime%5B1%5D=custom&status%5B0%5D=PASSED&status%5B1%5D=FAILED
```

the values must to be encoded, because param value is dynamic and can get from user input.
but the keys are mostly coded and in correct format for url.

that why option to encode only url params could be helpful.

any whay thank u about this util.
fill free to close this PR if you still disagree with me.

ljharb:
@elhay-av I hear you that the URLs look cleaner that way - but since both are received the same on the server, it doesn't sound like they _must_ have unencoded key brackets for any reason?

elhay-av:
It just look bad in browser url address field.

________________________________

ljharb:
@elhay-av this has now been addressed with the `encodeValuesOnly` option in #199. It will be in the next release.

Issue #173, opened 2016-10-04
Add support for RFC 1738


slavafomin:
Hello!

Thank you for this great module!

I'm using it to stringify API request query parameters and then generate MD5 hash of the resulted string. However, the server is expecting this string to be in RFC 1738 format, not in RFC 3986. This leads to incorrect hash, because spaces should be represented as `+`, not as `%20`.

It looks like server uses PHP's [http_build_query](http://php.net/manual/en/function.http-build-query.php) function to generate the query string and it uses RFC 1738 by default.

Could you please implement support for both RFC 1738 and RFC 3986 formats? It will make this library more compatible with different technologies.

Right now I have to use this workaround:

```js
let queryAsString = queryString
  .stringify(data)
  .replace(/%20/g, '+')
;
```

Thank you!

ljharb:
https://github.com/ljharb/qs/issues/152#issuecomment-211534354 seems relevant.

Note that you can already pass a custom `encoder` function to `stringify`, that will let you do the replacement there, and a custom "decoder" function to "parse".

slavafomin:
I would suggest this API:

```js
const qs = require('qs');

let result = qs.stringify(data, {
    format: qs.RFC1738,
//  format: qs.RFC3986
});
```

ljharb:
That's really not a bad idea at all.

Are there other RFCs that we'd want to support?

@slavafomin would you be willing to make a PR that adds these formats to both `parse` and `stringify`?

slavafomin:
@ljharb I would love to do so, but my time is extremely limited these days. I'm having a hard time supporting my own libraries right now. Sorry, I hope someone will have the time to implement this, because it looks like a good idea.

Max-Kolodezniy:
Done here. API interface is exactly the same as described above
https://github.com/ljharb/qs/pull/174

ljharb:
Done in #174.

slavafomin:
Great job!

PR #174, opened 2016-10-06
[New] Add RFC formatting support (#173)

Max-Kolodezniy:


lib/stringify.js
```diff
@@ -24,10 +24,16 @@ var defaults = {
         return toISO.call(date);
     },
     skipNulls: false,
-    strictNullHandling: false
+    strictNullHandling: false,
+    formatters : {
+        RFC3986 : function (value) { return value; },
+        /** @return {string} */
+        RFC1738 : function (value) { return value.toString().replace(/%20/g, '+'); }
```
ljharb: let's use `String(value)` here - also, what happens if `value` is null or undefined?

test/stringify.js
```diff
@@ -502,4 +502,25 @@ test('stringify()', function (t) {
...
+    });
+
+    t.test('RFC 3986 spaces serialization', function (st) {
+        st.equal(qs.stringify({ a: 'b c' }, { format: 'RFC3986' }), 'a=b%20c');
+        st.end();
+    });
+
+    t.test('Backward compatibility to RFC 3986 and edge cases', function (st) {
+        st.equal(qs.stringify({ a: 'b c' }, { format: 'UFO1234' }), 'a=b%20c');
+        st.equal(qs.stringify({ a: 'b c' }, { format: false }), 'a=b%20c');
```
ljharb: I'm wondering if it would be better to throw immediately if the provided format is invalid.

ljharb: Specifically, so if someone passes a specific RFC that we don't yet support, it doesn't silently pass.

Max-Kolodezniy: Agree, will fix that today

lib/stringify.js
Max-Kolodezniy: Roger that

Max-Kolodezniy:
Unfortunately, I cannot fix all Lint errors in this stringily without deep refactoring. I'd like to postpone it from now and change a little bit the .eslint file

slavafomin:
Great job, thanks!

However, I would suggest to pass format not as a string, but by referencing one of the predefined constants, like in [my example](https://github.com/ljharb/qs/issues/173#issuecomment-251311153).

We could put those under a property for namespacing, e.g. `qs.formats.RFC1738`.

Such approach would allow IDE's to highlight errors and to provide code completion when using the module.

Max-Kolodezniy:
You are right, IDE autocomplete support would be great. On my way to commit that.

lib/index.js
```diff
@@ -5,5 +5,9 @@ var parse = require('./parse');
 
 module.exports = {
     parse: parse,
+    // String constants
```
ljharb: let's group these under a `formats` object, so it's clear that "parse" isn't a format, and so the full list can be obtained via reflection.

lib/stringify.js
```diff
@@ -20,14 +20,20 @@ var defaults = {
     delimiter: '&',
     encode: true,
     encoder: utils.encode,
+    format: 'RFC3986',
+    formatters: {
+        /** @return {string} */
```
ljharb: please remove this; i don't want jsdoc anywhere near code that i have to maintain :-)

lib/stringify.js
```diff
@@ -43,9 +49,9 @@ var stringify = function stringify(object, prefix, generateArrayPrefix, strictNu
 
     if (typeof obj === 'string' || typeof obj === 'number' || typeof obj === 'boolean' || utils.isBuffer(obj)) {
         if (encoder) {
-            return [encoder(prefix) + '=' + encoder(obj)];
+            return [encoder(prefix) + '=' + formatter(encoder(obj))];
```
ljharb: does the formatter only need to apply to values, or should it also apply to keys?

lib/stringify.js
```diff
@@ -112,6 +120,12 @@ module.exports = function (object, opts) {
     var sort = typeof options.sort === 'function' ? options.sort : null;
     var allowDots = typeof options.allowDots === 'undefined' ? false : options.allowDots;
     var serializeDate = typeof options.serializeDate === 'function' ? options.serializeDate : defaults.serializeDate;
+    if (typeof options.format === 'undefined') {
+        options.format = defaults.format;
+    } else if (!(options.format in defaults.formatters)) {
+        throw new TypeError('Unknown format options provided. Valid values are RFC3986 and RFC1738.');
```
ljharb: Probably instead of a hardcoded list here, we should get the list off of `qs.formats`, or omit it entirely?

lib/stringify.js
slavafomin: What's wrong with jsdoc? It's very helpful with IDE code autocompletion and code proofing.

lib/stringify.js
ljharb: Sounds good!

Max-Kolodezniy: We have to notify user if provided format will not work. So, reaction to the wrong argument provided should be straight and self-explanatory

lib/stringify.js
Max-Kolodezniy: It's not common but possible. Will add that

lib/stringify.js
Max-Kolodezniy: Also, qs.formats is unavailable in stringify.js. And including index.js will cause sort of endless inclusion cycle

lib/index.js
```diff
@@ -4,6 +4,10 @@ var stringify = require('./stringify');
 var parse = require('./parse');
 
 module.exports = {
+    formats: {
```
ljharb: could this be stored in a separate file, and required in here? that way `stringify.js` could require it, and use the constant rather than the string

lib/stringify.js
```diff
@@ -43,9 +48,9 @@ var stringify = function stringify(object, prefix, generateArrayPrefix, strictNu
 
     if (typeof obj === 'string' || typeof obj === 'number' || typeof obj === 'boolean' || utils.isBuffer(obj)) {
         if (encoder) {
-            return [encoder(prefix) + '=' + encoder(obj)];
+            return [formatter(encoder(prefix)) + '=' + formatter(encoder(obj))];
         }
-        return [prefix + '=' + String(obj)];
+        return [formatter(prefix) + '=' + formatter(String(obj))];
```
ljharb: i'm wondering if we should pass a "key" and "value" string argument to `formatter` as well, in case the user wants to format them differently

lib/stringify.js
```diff
@@ -112,6 +119,12 @@ module.exports = function (object, opts) {
     var sort = typeof options.sort === 'function' ? options.sort : null;
     var allowDots = typeof options.allowDots === 'undefined' ? false : options.allowDots;
     var serializeDate = typeof options.serializeDate === 'function' ? options.serializeDate : defaults.serializeDate;
+    if (typeof options.format === 'undefined') {
+        options.format = defaults.format;
+    } else if (!(options.format in defaults.formatters)) {
```
ljharb: this will allow a format of `toString`, for example - instead of `in`, could we use `Object.prototype.hasOwnProperty.call`?

test/stringify.js
```diff
@@ -502,4 +502,35 @@ test('stringify()', function (t) {
...
+        st.equal(qs.stringify({ a: 'b c' }), 'a=b%20c');
+        st.end();
+    });
+
+    t.test('Edge cases and unknown formats', function (st) {
+        ['UFO1234', false, 1234, null, {}, []].forEach(
+            function (format) {
+                st.throws(
+                    function () {
+                        qs.stringify({ a: 'b c' }, {format: format});
```
ljharb: object curly braces need spaces inside them

lib/stringify.js
Max-Kolodezniy: I don't see how it can be formatted differently if user provide only one option. Formatter is dedicated function for string variables - keys and values and "=" sign and concatenation is not an formatter responsibility.

lib/index.js
Max-Kolodezniy: Hm, include the file for two strings gives uncleared and unjustified overhead. And in the same time - it replaces the only one occurrence in stringify.js : input parameter validation.
And mix constants and formatters to the one file will be out of KISS

test/stringify.js
Max-Kolodezniy: Fixed

lib/index.js
ljharb: What overhead? Modularity is far more valuable than avoiding a function call.

ljharb:
I think with the final change of moving the formats to a separate file, plus a rebase on top of latest master (ideally squashed down to a commit), this should be good to go.

Max-Kolodezniy:
I'm out of PC this week, travelling with smartphone only. Will do that in a
week

10 жовт. 2016 р. 9:12 пп "Jordan Harband" <notifications@github.com> пише:

lib/stringify.js
slavafomin: How could IDE parse the types of the variables, for example, when JavaScript doesn't have strict types?

Max-Kolodezniy:
This code has been changed. Now it's encapsulated in the separate file

12 жовт. 2016 р. 1:08 пп "Slava Fomin II" <notifications@github.com> пише:

ljharb:
@Max-Kolodezniy thanks! I think it's ready to go once it gets rebased down to a single commit. Since you're traveling, I'm happy to take care of it for you if you don't mind?

Max-Kolodezniy:
Feel free to do that!

12 жовт. 2016 р. 11:13 пп "Jordan Harband" <notifications@github.com> пише:

PR #175, opened 2016-10-15
[Tests] Turn on eslint for test files

sigorilla:
I noticed that you made [the mistake](https://github.com/ljharb/qs/commit/807c310924a897f1a4d1fa0588542a1ad6869fed#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R38) in the lint command. Because of this eslint did not check test folder.

### DIFF
--- a/test/.eslintrc
+++ b/test/.eslintrc
@@ -1,0 +1,9 @@
+{
+    "rules": {
+        "max-lines": 0,
+        "max-nested-callbacks": [2, 3],
+        "max-statements": 0,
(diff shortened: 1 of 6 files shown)


test/.eslintrc
```diff
@@ -0,0 +1,9 @@
+{
+	"rules": {
```
ljharb: although i like tabs, this repo is 4-spaced, so this file should be as well

sigorilla: Yep. I updated the files.

PR #187, opened 2016-12-19
Show benchmarking for array combine stuff

elidoran:
I made this PR separate from #185 because it's not my intent to add this stuff to master. Instead, I'm providing it just to mess with in its own branch.

Provides a new directory, `benchmark/`, with a script to benchmark different array combination implementations.

Adds dev dependencies:

1. benchmark
2. console.table
3. comma-number

Run benchmark via:

```sh
node benchmark/arraying.js
```

### DIFF
--- a/benchmark/arraying.js
+++ b/benchmark/arraying.js
@@ -1,0 +1,128 @@
+require('console.table')
+
+var comma = require('comma-number')
+  , Benchmark = require('benchmark')
+
+Benchmark.options.initCount  = 100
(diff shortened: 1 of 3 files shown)


benchmark/arraying.js
```diff
@@ -0,0 +1,128 @@
+require('console.table')
+
+var comma = require('comma-number')
+  , Benchmark = require('benchmark')
```
ljharb: comma last, please :-)

Issue #164, opened 2016-05-13
Query string don't contain empty object


aight8:
When I have the following object:
```JS
{
    myObject: {}
}
```

The query string don't contains the myObject.

ljharb:
What would you expect the query string to look like? In query strings, there's usually not "empty arrays" or "empty objects" - if it contains nothing, it usually doesn't exist.

ljharb:
Closing this, since this seems to be to be working as designed.

foisonocean:
@ljharb 
Hi, my case is empty objects may appears in an array, which will effect the length of the array.

```js
const obj1 = {query: [{foo: 'bar'}, {}, {bar: 'foo'}]};
console.log(obj1.query.length);
// 3

const query = qs.stringify(obj1);
// query: 'query%5B0%5D%5Bfoo%5D=bar&query%5B2%5D%5Bbar%5D=foo'


const obj2 = qs.parse(query);
// obj2: { query: [ { foo: 'bar' }, { bar: 'foo' } ] }

console.log(obj2.query.length);
// 2
```

In this example, the length is incorrect with the current implementation of `qs`. Is it possible to fix this?

ljharb:
I mean, you could fix it by using something other than empty objects to indicate nonexistence?

PR #193, opened 2017-01-06
add empty object and array values example.

up9cloud:
#164 has non-issue tag.

As my original expectation is `assert.equal(qs.stringify({ a: {} }), 'a');`

but actually is  `assert.equal(qs.stringify({ a: {} }), '');`

so I think it should be documented.

### DIFF
--- a/README.md
+++ b/README.md
@@ -278,6 +278,16 @@ Empty strings and null values will omit the value, but the equals sign (=) remai
 assert.equal(qs.stringify({ a: '' }), 'a=');
 ```
 
+Key with no values (such as an empty object or array) will return nothing:
+
+```javascript
+assert.equal(qs.stringify({ a: [] }), '');
+assert.equal(qs.stringify({ a: {} }), '');
+assert.equal(qs.stringify({ a: [{}] }), '');
+assert.equal(qs.stringify({ a: { b: []} }), '');
+assert.equal(qs.stringify({ a: { b: {}} }), '');
+```
+
 Properties that are set to `undefined` will be omitted entirely:
 
 ```javascript


up9cloud:
I'm not sure why. So i just wrote return nothing. XD

In fact, I wish there is a option for qs.stringify. (`keepKey = true`)

My opinion, there are 2 diff way to show the info.

1. more info

Server try to figure out what `'a='` is. (Sent from client by using `qs.stringify({a:null})`)

Know `a` exist, and value could be _empty string_ or _null_ or _undefined_.

2. less info

`''` recieved. (`qs.stringify({a:{}})`)

Server don't know `a` exist or not.

ljharb:
The query string spec means the server *can't ever* know if that key exists or not, because it doesn't exist - only keys with values exist. Query strings aren't a 1:1 analog to key/value pairs.

up9cloud:
I agree with you ("Query strings aren't a 1:1 analog to key/value pairs").
I'm just a little confused about "the value exist".
For me, null, undefined, empty object and array are same empty meaning here, 
but 2 kind of return value.
ps. I'm not a native speaker of English, so if there is any impolite things, my bad.
I tried to explain more clear.

ljharb:
`null` and `undefined` is a value that could result in the key's presence in the query string. However, an empty array isn't one - because "array" and "object" in query strings are modifiers on the *key*, not the value - so if the object or array is empty, there is no value to attach the key to.

up9cloud:
I think I got what you meant.
It may query data by condition `a is null` (null is a value), but it is no meaning to query by `a=[]` (no values).

how about
```
- Empty object and array values will return nothing:
+ Key with no values (empty object or array) will return nothing:
```

ljharb:
"such as an empty object or array" - sure, that sounds great

up9cloud:
i'd modify the description. Thank you!😊

PR #198, opened 2017-02-19
Encoder prefix boolean

joemcelroy:
Reason: We want to only encode our values to keep the url "pretty" help the url be shorter. 

For example: 
`{q: "test", categories: [["movie"],["Crime"]], actors: ["a", 'b-c', 'd & e', 'f=g'], writers: ['d'] }` 
to be `q=test&categories[0][0]=movie&categories[1][0]=Crime&actors[0]=a&actors[1]=b-c&actors[2]=d%20%26%20e&actors[3]=f%3Dg&writers[0]=d`
fully encoded: would escape `categories[0][0]`

welcome any feedback on improvements. This is a feature for searchkit, an open source ui framework for elasticsearch.

### DIFF
--- a/README.md
+++ b/README.md
@@ -228,8 +228,9 @@ assert.equal(unencoded, 'a[b]=c');
 This encoding can also be replaced by a custom encoding method set as `encoder` option:
 
 ```javascript
-var encoded = qs.stringify({ a: { b: 'c' } }, { encoder: function (str) {
+var encoded = qs.stringify({ a: { b: 'c' } }, { encoder: function (str, prefix) {
   // Passed in values `a`, `b`, `c`
+  // prefix is true for 'a' and 'b' values
   return // Return encoded string
 }})
 ```
@@ -400,7 +401,7 @@ assert.equal(nullsSkipped, 'a=b');
 
 ### Dealing with special character sets
 
-By default the encoding and decoding of characters is done in `utf-8`. If you 
+By default the encoding and decoding of characters is done in `utf-8`. If you
 wish to encode querystrings to a different character set (i.e.
 [Shift JIS](https://en.wikipedia.org/wiki/Shift_JIS)) you can use the
 [`qs-iconv`](https://github.com/martinheidegger/qs-iconv) library:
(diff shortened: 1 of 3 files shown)


ljharb:
Can you elaborate more on what this allows? Specifically, can you give an example of the default stringified version of your query string, compared to what you want it to look like?

joemcelroy:
sure, so we do not want to encode the keys, only the value. So an example in code
```
const obj = {q: "test", categories: [["movie"],["Crime"]], actors: ["a", 'b-c', 'd & e', 'f=g'], writers: ['d'] }
const str1 = qs.stringify(obj, { encode: true, encoder: (str, isPrefix) => {
    return isPrefix ? str : qsUtils.encode(str)
}})
const str2 = qs.stringify(obj, { encode: true })
```

str1 => `q=test&categories[0][0]=movie&categories[1][0]=Crime&actors[0]=a&actors[1]=b-c&actors[2]=d%20%26%20e&actors[3]=f%3Dg&writers[0]=d`
str2 => `q=test&categories%5B0%5D%5B0%5D=movie&categories%5B1%5D%5B0%5D=Crime&actors%5B0%5D=a&actors%5B1%5D=b-c&actors%5B2%5D=d%20%26%20e&actors%5B3%5D=f%3Dg&writers%5B0%5D=d`
 
We want str1. We use this string to serialize the state on the url for searchkit so needs to be 'readable' and short as possible.

ljharb:
(Just as an aside; what encoding are you doing that differs from the default?)

This seems similar to #171, which the user ended up closing (see https://github.com/ljharb/qs/pull/171#issuecomment-238275749) before we reached any decision.

It sounds like what you want is more "is it a key" or "is it a value", rather than "isPrefix" - is that accurate?

joemcelroy:
I use the default qs encoder. Yep my issue is the same as #171. Yep we could change it to key / value boolean.

ljharb:
If there was an option to only encode values, not keys (as opposed to changing the `encoder` API), would that also meet your needs?

joemcelroy:
yep that would meet my needs. I can submit a PR for that if you want?

ljharb:
Let's do that - add a `stringify` option for `doNotEncodeKeys` that defaults to `false`, and when it is `true`, it skips the `encoder` call entirely for anything that's a key.

joemcelroy:
cool yep that sounds good. Il do that now.

PR #199, opened 2017-02-19
encodeValuesOnly option

joemcelroy:
linked to #198 and #171 

stringify option for doNotEncodeKeys that defaults to false, and when it is true, it skips the encoder call entirely for anything that's a key.

### DIFF
--- a/lib/stringify.js
+++ b/lib/stringify.js
@@ -21,6 +21,7 @@ var defaults = {
     delimiter: '&',
     encode: true,
     encoder: utils.encode,
+    encodeValuesOnly: false,
     serializeDate: function serializeDate(date) { // eslint-disable-line func-name-matching
         return toISO.call(date);
     },
@@ -39,7 +40,8 @@ var stringify = function stringify( // eslint-disable-line func-name-matching
     sort,
     allowDots,
     serializeDate,
-    formatter
+    formatter,
+    encodeValuesOnly
 ) {
     var obj = object;
     if (typeof filter === 'function') {
@@ -48,7 +50,7 @@ var stringify = function stringify( // eslint-disable-line func-name-matching
         obj = serializeDate(obj);
     } else if (obj === null) {
         if (strictNullHandling) {
-            return encoder ? encoder(prefix) : prefix;
+            return encoder && !encodeValuesOnly ? encoder(prefix) : prefix;
         }
 
         obj = '';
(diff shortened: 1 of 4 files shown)


joemcelroy:
il pick this up in the morning if any issues!

test/stringify.js
```diff
@@ -535,4 +535,18 @@ test('stringify()', function (t) {
         );
         st.end();
     });
+
+    t.test('doNotEncodeKeys', function (st) {
+        st.equal(qs.stringify({ a: 'b', c: ['d', 'e=f'], f: [['g'], ['h']] }, { doNotEncodeKeys: true })
+        , 'a=b&c[0]=d&c[1]=e%3Df&f[0][0]=g&f[1][0]=h');
```
ljharb: comma last, please.

test/stringify.js
```diff
@@ -535,4 +535,18 @@ test('stringify()', function (t) {
...
+        st.equal(qs.stringify({ a: 'b', c: ['d', 'e'], f: [['g'], ['h']] })
```
ljharb: if the entire function invocation can't fit on one line, then every argument must be on a line by itself, as well as the closing `)`.

lib/stringify.js
```diff
@@ -19,6 +19,7 @@ var toISO = Date.prototype.toISOString;
 
 var defaults = {
     delimiter: '&',
+    doNotEncodeKeys: false,
```
ljharb: I'm not sure if i like this name better, or `encodeValuesOnly`. Thoughts?

joemcelroy: I agree, encodeValuesOnly is better. I don't like negative option names. I'll update it to that

ljharb: Yeah, I strongly feel that "absence" and `false` must always do the same thing when possible (some of the options in qs predate me) but it does sometimes end up with awkward option names.

test/stringify.js
```diff
@@ -535,4 +535,33 @@ test('stringify()', function (t) {
         );
         st.end();
     });
+
+    t.test('doNotEncodeKeys', function (st) {
```
ljharb: "encodeValuesOnly"?

test/stringify.js
```diff
@@ -535,4 +535,33 @@ test('stringify()', function (t) {
...
+        st.equal(
+            qs.stringify(
+                { a: 'b', c: ['d', 'e'], f: [['g'], ['h']] }
+            ),
+            'a=b&c%5B0%5D=d&c%5B1%5D=e&f%5B0%5D%5B0%5D=g&f%5B1%5D%5B0%5D=h'
+        );
+        st.end();
+    });
+
+    t.test('doNotEncodeKeys - strictNullHandling', function (st) {
```
ljharb: "encodeValuesOnly"

joemcelroy:
thanks good spot, my morning eyes missed it! fixed and squashed. Hopefully its good now.

PR #201, opened 2017-03-02
fix: follow allowPrototypes during merge

### LINKED ISSUE #200: prototype override protection bypass problem still exists.
as the bug https://snyk.io/vuln/npm:qs:20170213 report fixed, but the other bypass

```
a = qs.parse("[=toString", {allowPrototypes: false})
// { toString: true }
```

still exists.

dead-horse:
closes #200

### DIFF
--- a/test/parse.js
+++ b/test/parse.js
@@ -152,8 +152,6 @@ test('parse()', function (t) {
         st.end();
     });
 
-    t.deepEqual(qs.parse('a[b]=c&a=d'), { a: { b: 'c', d: true } }, 'can add keys to objects');
-
     t.test('correctly prunes undefined values when converting an array to an object', function (st) {
         st.deepEqual(qs.parse('a[2]=b&a[99999999]=c'), { a: { 2: 'b', 99999999: 'c' } });
         st.end();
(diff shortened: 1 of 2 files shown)


dead-horse:
I'm not sure what's the purpose of this commit https://github.com/ljharb/qs/commit/50ea161bb0711b6d82d0215a4553d9181bb91f63 , but I think it didn't fix prototype overwrite.

ljharb:
The purpose of the commit is in the commit message, and the added tests shows exactly what it inarguably did fix.

test/parse.js
```diff
@@ -444,6 +442,44 @@ test('parse()', function (t) {
...
+
+        st.deepEqual(
+            qs.parse('a[b]=c&a=toString'),
+            { a: { b: 'c' } },
+            'can not overwrite prototype'
+        );
+
+        st.deepEqual(
+            qs.parse('a[b]=c&a=toString', { allowPrototypes: true }),
+            { a: { b: 'c', toString: true } },
```
ljharb: in this case, shouldn't `a[b]` be entirely overwritten by the string "toString"?

dist/qs.js
```diff
@@ -584,4 +584,4 @@ exports.isBuffer = function (obj) {
 };
 
 },{}]},{},[2])(2)
-});
\ No newline at end of file
+});
```
ljharb: this file shouldn't be updated in PRs; please revert it.

test/parse.js
dead-horse: I found this test case `t.deepEqual(qs.parse('a[b]=c&a=d'), { a: { b: 'c', d: true } }, 'can add keys to objects');`, is this a feature by design?

dead-horse:
after 50ea161 ,

```js
qs.parse('[=a'); // { a: true }
qs.parse(']=a'); // { ']': 'a'}
```

Is this the expected result?

dist/qs.js
dead-horse: fixed

dead-horse:
if we revert the change of `var parent =  /^([^[\]]*)/;` to `var parent = /^([^\[\]]*)/;`, the only failing test case is:

```
not ok 135 should be equivalent
  ---
    operator: deepEqual
    expected: |-
      { ']': 'toString' }
    actual: |-
      {}
    at: Test.<anonymous> (/Users/deadhorse/git/github.com/ljharb/qs/test/parse.js:441:12)
  ...
```

ljharb:
@dead-horse the second line in your example is the expected result, which is what it did, actually, fix. The first line is an oversight, obviously, hence #200.

ljharb:
That test is valid; I would expect `[=toString` to become `{ '[': 'toString' }` as well.

test/parse.js
ljharb: good point :-) let's keep this as-is for now.

popomore:
I want to know the diff in https://github.com/ljharb/qs/commit/50ea161bb0711b6d82d0215a4553d9181bb91f63?diff=split#diff-e6c9a6bca996bc454cc244d17bfeda5cL125

```diff
- st.deepEqual(qs.parse('a[]=b&a[t]=u&a[hasOwnProperty]=c', { allowPrototypes: false }), { a: { '0': 'b', c: true, t: 'u' } });
+ st.deepEqual(qs.parse('a[]=b&a[t]=u&a[hasOwnProperty]=c', { allowPrototypes: false }), { a: { 0: 'b', t: 'u' } });
```

Is it a expected change? it seems a break change that change string to nuber.

ljharb:
@popomore in JS, all keys are strings (or Symbols), so `{ 0: true }` and `{ '0': true }` are identical and indistinguishable.

PR #213, opened 2017-06-13
[New] Prefix option for parse function

dlwalsh:
The primary motivation for this change is that react-router v4 returns the `location.search` property with a leading question mark. Whilst it's trivial to remove this, it's nicer still to have the library handle it.

### DIFF
--- a/lib/parse.js
+++ b/lib/parse.js
@@ -18,7 +18,9 @@ var defaults = {
 
 var parseValues = function parseQueryStringValues(str, options) {
     var obj = {};
-    var parts = str.split(options.delimiter, options.parameterLimit === Infinity ? undefined : options.parameterLimit);
+    var cleanStr = options.ignoreQueryPrefix ? str.replace(/^\?/, '') : str;
+    var limit = options.parameterLimit === Infinity ? undefined : options.parameterLimit;
+    var parts = cleanStr.split(options.delimiter, limit);
 
     for (var i = 0; i < parts.length; ++i) {
         var part = parts[i];
@@ -136,6 +138,7 @@ module.exports = function (str, opts) {
         throw new TypeError('Decoder has to be a function.');
     }
 
+    options.ignoreQueryPrefix = options.ignoreQueryPrefix === true;
     options.delimiter = typeof options.delimiter === 'string' || utils.isRegExp(options.delimiter) ? options.delimiter : defaults.delimiter;
     options.depth = typeof options.depth === 'number' ? options.depth : defaults.depth;
     options.arrayLimit = typeof options.arrayLimit === 'number' ? options.arrayLimit : defaults.arrayLimit;
--- a/test/stringify.js
+++ b/test/stringify.js
@@ -18,6 +18,11 @@ test('stringify()', function (t) {
         st.end();
     });
 
+    t.test('adds query prefix', function (st) {
+        st.equal(qs.stringify({ a: 'b' }, { addQueryPrefix: true }), '?a=b');
+        st.end();
+    });
+
     t.test('stringifies a nested object', function (st) {
         st.equal(qs.stringify({ a: { b: 'c' } }), 'a%5Bb%5D=c');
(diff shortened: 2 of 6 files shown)


dlwalsh:
I also contemplated making `prefix` either a string or a regular expression, as with `delimiter`. But I was unsure of the merits of regex in this instance. Thoughts?

lib/parse.js
```diff
@@ -13,12 +13,15 @@ var defaults = {
...
     parameterLimit: 1000,
     plainObjects: false,
+    prefix: '',
     strictNullHandling: false
 };
 
 var parseValues = function parseQueryStringValues(str, options) {
     var obj = {};
-    var parts = str.split(options.delimiter, options.parameterLimit === Infinity ? undefined : options.parameterLimit);
+    var cleanStr = str.indexOf(options.prefix) === 0 ? str.substr(options.prefix.length) : str;
```
ljharb: please never use `substr`; it's either nonstandard or deprecated. Always use `slice`.

dlwalsh: Fixed.

dlwalsh:
No particular use case for a generic prefix, other than my inclination towards a flexible API.

I'm happy to change it to a boolean option that would simply trim the leading `?`. In that case, what should it be called? Perhaps `ignoreQueryPrefix`?

ljharb:
I agree that your API is properly flexible :-) I think `ignoreQueryPrefix` might be simpler though, for this use case.

We might as well add an `addQueryPrefix` option to `parse` as well, so they can be symmetrical.

dlwalsh:
I've updated parse to use the `ignoreQueryPrefix` option.

I've tried to add `addQueryPrefix` to stringify. However, even the slightest change to this function throws complexity errors in the tests. It seems to me that stringify is in need of a more generic refactor.

lib/parse.js
```diff
@@ -18,7 +18,9 @@ var defaults = {
 
 var parseValues = function parseQueryStringValues(str, options) {
     var obj = {};
-    var parts = str.split(options.delimiter, options.parameterLimit === Infinity ? undefined : options.parameterLimit);
+    var cleanStr = options.ignoreQueryPrefix && str.slice(0, 1) === '?' ? str.slice(1) : str;
```
ljharb: `str.charAt(0) === '?'` might be cleaner than the slice(0, 1) here - alternatively, `options.ignoreQueryPrefix ? str.replace(/^\?/, '') : str`?

Also, please parenthesize the ternary conditional since there's multiple items.

dlwalsh: Done.

dlwalsh:
I've cleaned up parse as requested. I've also added `addQueryPrefix` to stringify, with some modifications to .eslintrc.

robcalcroft:
What version of `qs` will have this feature as the current NPM release doesn't?

shri3k:
^would like to know this as well.

ljharb:
Whatever the next one is. This is a minor, so it'll be in the next release that a proper ^ range covers.

test/stringify.js
```diff
@@ -18,6 +18,11 @@ test('stringify()', function (t) {
         st.end();
     });
 
+    t.test('adds query prefix', function (st) {
+        st.equal(qs.stringify({ a: 'b' }, { addQueryPrefix: true }), '?a=b');
+        st.end();
+    });
```
nerfologist: Hi, regarding the new `addQueryPrefix` option, I noticed the following situation:
```
> qs.stringify({ a: 1, b: 2}, { addQueryPrefix: true })
'?a=1&b=2'
> qs.stringify({}, { addQueryPrefix: true})
'?'
```
In my use case I don't want to keep the lonely `?` if I'm stringifying an empty object.

What do you think about it?

ljharb: That seems like a reasonable modification - I haven't released the feature yet, so can you get a PR up asap for that?

ljharb: Done in #217, thanks!

PR #217, opened 2017-06-27
Handle stringifying empty objects with addQueryPrefix

nerfologist:
Submitting PR as agreed, this should fix the issue described in #213:

Stringifying an empty object would produce a string such as '?'. With
this fix, we output an empty string instead.

Sorry about increasing complexity and max-statements in `.eslintrc`,
I didn't see a quick way to reduce complexity.

Thanks for this great library 👍

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -4,12 +4,12 @@
     "extends": "@ljharb",
 
     "rules": {
-        "complexity": [2, 27],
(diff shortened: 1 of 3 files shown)


ljharb:
(This is a patch because the original feature has never been released; otherwise, it'd be a semver-major)

PR #224, opened 2017-09-09
Fix parsing & compacting very deep objects

dougwilson:
This is a change of the internal `compact` utility to use an iterative approach instead of recursion in order to better handle very, very deep objects without a stack overflow. It was noticed on CharkaCore because it has a much smaller maximum stack than v8 does, though it can still happen on v8 (which the test does).

/cc @kunalspathak

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -11,6 +11,29 @@ var hexTable = (function () {
     return array;
 }());
 
+var compactQueue = function (queue) {
+    var obj;
+
+    while (queue.length) {
+        var item = queue.pop();
+        obj = item.obj[item.prop];
+
+        if (Array.isArray(obj)) {
+            var compacted = [];
+
+            for (var j = 0; j < obj.length; ++j) {
+                if (typeof obj[j] !== 'undefined') {
+                    compacted.push(obj[j]);
+                }
+            }
+
+            item.obj[item.prop] = compacted;
+        }
+    }
+
+    return obj;
+};
+
 exports.arrayToObject = function (source, options) {
     var obj = options && options.plainObjects ? Object.create(null) : {};
     for (var i = 0; i < source.length; ++i) {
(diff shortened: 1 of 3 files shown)


lib/utils.js
```diff
@@ -144,38 +144,48 @@ exports.encode = function (str) {
...
 
-    var refs = references || [];
-    var lookup = refs.indexOf(obj);
-    if (lookup !== -1) {
-        return refs[lookup];
-    }
+    return exports.compactQueue(queue);
+};
+
+exports.compactQueue = function (queue) {
```
ljharb: does this need to be publicly exported? It'd be nice not to increase the public API.

dougwilson: It doesn't, but didn't see the `utils` object being exported out through the `index` entry point. I wasn't sure if adding to `exports` was just the style or not (most of it was changed to the Hapi project style when it lived there for a bit and I just wasn't super familiar).

TL;DR thought that it would fit in that way, but I'll in-export the function 👍

ljharb: It's certainly consistent :-) but I'm in a slow yet determined march to move away from the hapi style, and minimize the publicly reachable API of `qs`, so the change is appreciated!

dougwilson: haha, making the change is no problem at all :) I try to copy the existing code style as much as possible to blend it in, so wasn't sure, hehe

ljharb:
Looks like your new test is only failing in Turbofan-enabled nodes - 8.3+.

dougwilson:
Yep, just saw that and taking a look :)

dougwilson:
Ok, for whatever reason Turbofan was just failing to collapse in the same function Crankshaft was doing in the actual parsing. So now parsing itself is iterative instead of recursive. Probably even a better win overall now for everyone.

ljharb:
Released in v6.5.1

Issue #223, opened 2017-08-18
Feature request: option for deciding how to parse query without argument


timhwang21:
Hi,

I was wondering if it would be possible to support passing an option for deciding how to parse queries without arguments? Currently `url?foo` can be parsed as either the empty string or as `null`. I am wondering if it's possible to have the parser treat a query without arguments as `true`.

Some background: I ran into a case in our codebase which does this, it seemed odd to me but apparently there's nothing in [the standards](https://tools.ietf.org/html/rfc3986#section-3.4) that forbid queries without arguments. Are there more specific standards (e.g. the node implementation) that this library is trying to adhere to?

Thanks!

ljharb:
I'm not sure what you mean by "without arguments" - in `url?foo`, `foo` is the query string; `qs.parse('foo&bar=')` indeed returns `{ bar: "", foo: "" }`, and `qs.stringify({ foo: '', bar: true }))` returns `'foo=&bar=true'`.

You can certainly do this, however:
```js
qs.stringify({ foo: '', bar: null }, { strictNullHandling: true }); // 'foo&bar='
qs.parse('foo&bar=', { strictNullHandling: true }); // { bar: '', foo: null }
```

Does that cover your use case?

timhwang21:
"Query string field without a value" would probably be more accurate, sorry.

I was mainly wondering if something like this were possible -- basically, supporting valueless fields being parsed as having a given value:

```javascript
qs.parse('foo&bar=', { parseEmptyValueAs: true }); // { bar: '', foo: true }
// possibly also useful
qs.parse('foo&bar=', { parseEmptyValueAs: '123' }); // { bar: '', foo: '123' }
```

Thanks!

ljharb:
That's interesting. Effectively, it'd be an option for both parse and stringify that defined what primitive "an absent value" was - so `qs.stringify({ foo: 123, bar: 456 }, { emptyValue: 123 })` would give `'foo&bar=456'`.

I think that's reasonable, since it's pretty cumbersome to do that otherwise.

timhwang21:
Sounds good, do you think it would make sense to (eventually) combine with `strictNullHandling`? Instead of `true` that option can just take whatever value you want the absent value to take. Otherwise the user may get unexpected results with `{ strictNullHandling: true, emptyValue: 123 }` -- it'd be user error for sure, but can be reduced by consolidating the settings.

ljharb:
[...]

PR #226, opened 2017-09-16
Feature> emptyValue options for parse & stringify

timhwang21:
Addresses #223.

Behavior:
* When parsing and encountering empty values, replaces them with
opt.emptyValue (default to '')
* When stringifying and encountering value === opt.emptyValue, outputs a key
without a value

### DIFF
--- a/lib/parse.js
+++ b/lib/parse.js
@@ -15,6 +15,7 @@ var defaults = {
     decoder: utils.decode,
     delimiter: '&',
     depth: 5,
+    emptyValue: '',
     ignoreQueryPrefix: false,
     interpretNumericEntities: false,
     parameterLimit: 1000,
@@ -82,7 +83,7 @@ var parseValues = function parseQueryStringValues(str, options) {
         var key, val;
         if (pos === -1) {
             key = options.decoder(part, defaults.decoder, charset, 'key');
-            val = options.strictNullHandling ? null : '';
+            val = options.strictNullHandling ? null : options.emptyValue;
         } else {
             key = options.decoder(part.slice(0, pos), defaults.decoder, charset, 'key');
             val = utils.maybeMap(
@@ -212,6 +213,11 @@ var normalizeParseOptions = function normalizeParseOptions(opts) {
     if (typeof opts.charset !== 'undefined' && opts.charset !== 'utf-8' && opts.charset !== 'iso-8859-1') {
         throw new TypeError('The charset option must be either utf-8, iso-8859-1, or undefined');
     }
+
+    if (opts.strictNullHandling === true && has.call(opts, 'emptyValue')) {
+        throw new TypeError('Passing true for strictNullHandling will cause emptyValue to be ignored');
+    }
+
     var charset = typeof opts.charset === 'undefined' ? defaults.charset : opts.charset;
 
     return {
@@ -225,6 +231,7 @@ var normalizeParseOptions = function normalizeParseOptions(opts) {
         delimiter: typeof opts.delimiter === 'string' || utils.isRegExp(opts.delimiter) ? opts.delimiter : defaults.delimiter,
         // eslint-disable-next-line no-implicit-coercion, no-extra-parens
         depth: (typeof opts.depth === 'number' || opts.depth === false) ? +opts.depth : defaults.depth,
+        emptyValue: has.call(opts, 'emptyValue') ? defaults.emptyValue : opts.emptyValue,
         ignoreQueryPrefix: opts.ignoreQueryPrefix === true,
         interpretNumericEntities: typeof opts.interpretNumericEntities === 'boolean' ? opts.interpretNumericEntities : defaults.interpretNumericEntities,
         parameterLimit: typeof opts.parameterLimit === 'number' ? opts.parameterLimit : defaults.parameterLimit,
--- a/lib/parse.js
+++ b/lib/parse.js
@@ -20,8 +20,7 @@ var defaults = {
     interpretNumericEntities: false,
     parameterLimit: 1000,
     parseArrays: true,
-    plainObjects: false,
-    strictNullHandling: false
+    plainObjects: false
 };
 
 var interpretNumericEntities = function (str) {
@@ -83,7 +82,7 @@ var parseValues = function parseQueryStringValues(str, options) {
         var key, val;
         if (pos === -1) {
             key = options.decoder(part, defaults.decoder, charset, 'key');
-            val = options.strictNullHandling ? null : options.emptyValue;
+            val = options.emptyValue;
         } else {
             key = options.decoder(part.slice(0, pos), defaults.decoder, charset, 'key');
             val = utils.maybeMap(
@@ -214,9 +213,7 @@ var normalizeParseOptions = function normalizeParseOptions(opts) {
         throw new TypeError('The charset option must be either utf-8, iso-8859-1, or undefined');
     }
 
-    if (opts.strictNullHandling === true && has.call(opts, 'emptyValue')) {
-        throw new TypeError('Passing true for strictNullHandling will cause emptyValue to be ignored');
-    }
+    var defaultEmptyValue = opts.strictNullHandling === true ? null : defaults.emptyValue
 
     var charset = typeof opts.charset === 'undefined' ? defaults.charset : opts.charset;
 
@@ -231,13 +228,12 @@ var normalizeParseOptions = function normalizeParseOptions(opts) {
         delimiter: typeof opts.delimiter === 'string' || utils.isRegExp(opts.delimiter) ? opts.delimiter : defaults.delimiter,
         // eslint-disable-next-line no-implicit-coercion, no-extra-parens
         depth: (typeof opts.depth === 'number' || opts.depth === false) ? +opts.depth : defaults.depth,
-        emptyValue: has.call(opts, 'emptyValue') ? defaults.emptyValue : opts.emptyValue,
+        emptyValue: has.call(opts, 'emptyValue') ? opts.emptyValue : defaultEmptyValue,
         ignoreQueryPrefix: opts.ignoreQueryPrefix === true,
         interpretNumericEntities: typeof opts.interpretNumericEntities === 'boolean' ? opts.interpretNumericEntities : defaults.interpretNumericEntities,
         parameterLimit: typeof opts.parameterLimit === 'number' ? opts.parameterLimit : defaults.parameterLimit,
         parseArrays: opts.parseArrays !== false,
         plainObjects: typeof opts.plainObjects === 'boolean' ? opts.plainObjects : defaults.plainObjects,
-        strictNullHandling: typeof opts.strictNullHandling === 'boolean' ? opts.strictNullHandling : defaults.strictNullHandling
     };
 };
 
--- a/lib/stringify.js
+++ b/lib/stringify.js
@@ -32,6 +32,7 @@ var defaults = {
     charset: 'utf-8',
     charsetSentinel: false,
     delimiter: '&',
+    emptyValue: null,
     encode: true,
     encoder: utils.encode,
     encodeValuesOnly: false,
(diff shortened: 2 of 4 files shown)


lib/stringify.js
```diff
@@ -27,13 +29,15 @@ var defaults = {
     },
     skipNulls: false,
     strictNullHandling: false
+    // emptyValue: some flag
```
timhwang21: I'm not sure what the best way to handle this case is, as there needs to way to recognize "user provided no option" so the default behavior of outputting `foo=` as opposed to `foo` can be used.

I've seen things like using the global `Infinity`, storing the option in an object and using `hasOwnProperty`, etc. among others but not sure what you'd prefer. Thanks!

lib/parse.js
```diff
@@ -142,6 +143,11 @@ module.exports = function (str, opts) {
         throw new TypeError('Decoder has to be a function.');
     }
 
+    if (options.strictNullHandling === true && has.call(options, 'emptyValue')) {
+        throw new TypeError('strictNullHandling takes precedence over emptyValue.');
+        // TODO: I don't think this has to throw an error, but console.warn is a lint error
```
timhwang21: Also not sure what the best way to handle this would be.

lib/stringify.js
```diff
@@ -131,6 +140,7 @@ module.exports = function (object, opts) {
 
     var delimiter = typeof options.delimiter === 'undefined' ? defaults.delimiter : options.delimiter;
     var strictNullHandling = typeof options.strictNullHandling === 'boolean' ? options.strictNullHandling : defaults.strictNullHandling;
+    var emptyValue = has.call(options, 'emptyValue') ? options.emptyValue : 'SOME_FLAG'; // TODO
```
timhwang21: This causes a lint error for `complexity` and `max-statements`, how should I resolve these? Thanks.

ljharb: Just bump the thresholds in eslintrc til it passes ¯\\\_(ツ)_/¯ those rules aren't particularly important

lib/stringify.js
timhwang21: Yeah -- let's say we don't provide a default. This means when stringifying `{ foo: undefined }` we get `'foo'` whereas before we get `''` (line 77). If we do provide a default, stringifying `{ foo: [whatever our chosen default is]}` gives `'foo'` instead of whatever it gave before my change. So without some more involved code, stringifying whatever value we choose as the default would change from whatever it was before to `'foo'`.

ljharb: I'd like this not to be a breaking change, so `qs.stringify({ foo: undefined })` needs to return the same value in master as it does in this PR.

ljharb:
@timhwang21 is this still something you're interested in?

timhwang21:
@ljharb sure! Sorry this fell off my radar as we ended up solving the problem we were facing on our end.

lib/stringify.js
```diff
@@ -63,8 +67,8 @@ var stringify = function stringify( // eslint-disable-line func-name-matching
         obj = serializeDate(obj);
     }
 
-    if (obj === null) {
-        if (strictNullHandling) {
+    if (obj === null || obj === emptyValue) {
+        if (strictNullHandling || (obj === emptyValue && emptyValue !== null)) {
             return encoder && !encodeValuesOnly ? encoder(prefix, defaults.encoder, charset) : prefix;
```
timhwang21: This code is sort of confusing, but I think this results in the most expected outcomes:

1. No options: `{ a: null }` becomes `'a='`
2. `strictNullHandling`: `{ a: null }` becomes `'a'`
3. `emptyValue: 'something'`: `{ a: 'something' }` becomes `'a'`
4. `emptyValue: null`: `{ a: null }` becomes `'a'`

The relevant cases for this patch are 3 and 4. Without the additional check on line 71, case 3 would give `'a='` and case 4 would give `'a='`. However to me it seems if the user provides `emptyValue` they implicitly want "strict empty handling" to be true, as I don't think returning `'a='` for 3 and 4 is ever a useful outcome.

This also means that `emptyValue: null` is equivalent to `strictNullHandling: true`.

lib/parse.js
```diff
@@ -187,6 +188,10 @@ module.exports = function (str, opts) {
         throw new TypeError('Decoder has to be a function.');
     }
 
+    if (options.strictNullHandling === true && has.call(options, 'emptyValue')) {
```
ljharb: we check truthiness of `strictNullHandling` elsewhere; why `=== true` here?

lib/stringify.js
ljharb: Hmm, I'm not sure I agree - `a=` explicitly means "empty" to me, whereas `a` could be shorthand for `a: true`. I think I might expect case 3 to become `a=`, and case 6 (`emptyValue: 'something' + `strictNullHandling` makes `{ a: 'something' }` become `a`?

lib/parse.js
timhwang21: I think the previous conversation was lost on rebase, we had said that it doesn't really make sense for `strictNullHandling` and `emptyValue` to be used together, since both try to determine what the "empty" value is.

lib/stringify.js
timhwang21: I do agree that case 3 giving `a=` is more expected, but I think it's less useful. I'm imagining the following:

1. User does `qs.parse('foo&bar=1', { emptyValue: true })` and gets `{ foo: true, bar: 1 }`
2. User increments `bar` by 1 and re-serializes: `qs.stringify({ foo: true, bar: 2 }, { emptyValue: true })`. I'd expect to get `foo&bar=2` which follows the format of the original input, not `foo=&bar=2` here.

Regardless... I do agree that requiring `strictNullHandling: true` for this behavior for `stringify` _does_ feel more logical. Returning `a` just strikes me as more in line with the use case this PR is meant to address.

lib/parse.js
ljharb: oh sure, i meant why not this:

```suggestion
    if (options.strictNullHandling && has.call(options, 'emptyValue')) {
```

timhwang21: Ah, gotcha. Thanks for explaining!

lib/stringify.js
ljharb: I guess the confusion is because we have an implicit empty value of null, and a separate “strict null handling” var. if it was “strict empty value handling”, then it’d be easy to make it make sense. I’m not sure what the best approach here is.

ljharb: I think it makes sense to forbid the strictNullHandling option entirely when emptyValue is not null.

However, then it may make sense to add a new option, strictEmptyHandling, that subsumes strictNullHandling in all cases?

timhwang21: +1 on the first part, that makes it more consistent with `parse` behavior.

I'm honestly still unconvinced on the second part, as I still don't think anyone would ever want to use `stringify` + `emptyValue` without also wanting "strict" behavior. It just seems like it would lead to user error since the desired outcome requires `emptyValue` to always be used in tandem with `strictFooHandling`...

ljharb: Both with and without the strict option, what’s the roundtrip behavior of a null?

timhwang21:
```javascript
stringify({ a: null })
// => 'a='
parse('a=')
// => { a: '' }

// strict - matches
stringify({ a: null }, { strictNullHandling: true })
// => 'a'
parse('a', { strictNullHandling: true })
// => { a: null }
```

I'd expect `emptyValue: 'foobar'` to basically be `strictFoobarHandling: true`:

```javascript
stringify({ a: 'foobar' }, { emptyValue: 'foobar' })
// => 'a'
parse('a', { emptyValue: 'foobar' })
// => { a: 'foobar' }
```

issuefiler:
I suggest the following.

```
INPUT
OUTPUT

""
{}

"a"
{a: null}

"a="
{a: ""}

"a=null"
{a: "null"}

"a[0]" (arrayFormat: "indices")
{a: []}

"a[0]=" (arrayFormat: "indices")
{a: [""]}

"a[]" (arrayFormat: "brackets")
{a: []}

"a[]=" (arrayFormat: "brackets")
{a: [""]}

{a: undefined}
""

{a: null}
"a"

{a: ""}
"a="

{a: "null"}
"a=null"
```

-----

I believe interpreting the presense of a key as its value being truthy is up to the users.
`qs` should, by default, just focus on conveying _what keys are “defined”_ and _what values, if any, they hold_.
`qs` doesn’t know whether `"a"` is boolean or not.

If how others interpret the stringified is concerned, why don’t you add “_Do not convey (do omit) the presense of `null`ed keys_” and “_Do not convey (do omit) the presense of empty arrays_” options?

jayaddison:
Just adding a note that I'm interested in this functionality too.

While I mostly agree with @issuefiler's desired behaviour (`a` is null, `a=` is an empty string, 'undefined' implies key-not-present, ...), I also agree with the suggestion that this should be an option via `options.emptyValue`.

Given that truthy/falsy checks can be tricky and non-obvious to reason about, it's worth being careful to reduce impact to existing codebases (especially since there are many of them) when they upgrade `qs`.

lib/stringify.js
ljharb: cc @timhwang21, have you had any time to work on this PR?

timhwang21: In all honesty, no, this entirely fell off my radar. Will catch up on the past few years of discussion 😅

timhwang21:
@ljharb after thinking about this some more, I'm liking the idea of `strictEmptyHandling` (and `skipEmpty` as well) you suggested. I'm proposing the following API:

Existing options:
- `strictNullHandling` -- Now no longer does anything by itself. When set to `true`, sets `strictEmptyHandling` to true and sets `emptyValues` to `null`.
- `skipNulls` -- Now no longer does anything by itself. When set to `true`, sets `skipEmpty` to true and sets `emptyValues` to `null`.

New options:
- `emptyValues` -- Defaults to `null`. Represents the value `qs` treats as "empty."
- `strictEmptyHandling`
  - `parse` -- when set to `true` and given a value without a trailing `=`, parses as `emptyValue` instead of `''`.
  - `stringify` -- when set to `true` and given a k-v pair with where value `=== emptyValue`, returns a the key without an equals sign.
- `skipEmpty`
  - ~`parse`~ -- not implemented
  - `stringify` -- when set to `true` and stringifying a k-v pair where value `=== emptyValue`, returns the empty string.

Edge cases:
- When stringifying, we cannot treat multiple values as empty. We can allow passing an array `emptyValues` to `stringify`; however, this breaks the parallel options setup between `stringify` and `parse`, because multiple empty values don't make sense for `parse`. The use case would be that when stringifying, we can do things like treat both `null` and `''` as empty. Alternatively, we can stay with the API of the current PR and keep the concept of empty and strict null entirely separate; however, I feel this makes the API harder to understand.
- `strictNullHandling: true` or `skipNulls: true` AND `emptyValue != null`. The conflict arises from `strictNullHandling` wanting to set `emptyValue` to `null`. We can either have this throw an error like my current PR does or have the user-specified `emptyValue` override `null`. I think the latter is reasonable but confusing.

Summary:
- If not passing any value to `emptyValue` (meaning the default of `null` is used), `strict{Null,Empty}Handling` and `skip{Nulls,Empty}` are equivalent.
- Existing code that doesn't specify the new options will see no change in behavior.
- I think we should choose one of the two options I proposed above for handling the need to serialize BOTH `null` an `''` as an empty value. This seems like a very reasonable and desirable use case. I personally favor `options.emptyValue: any` for `parse` and `options.emptyValues: any[]` for `stringify`.

Please let me know what you think. Thanks!

ljharb:
@timhwang21 I like your new API suggestion. I like throwing an error when conceptually conflicting options are passed.

I'm not sure I understand the last bullet point in the "Summary", but if you're able to update the PR, maybe it will become clear to me :-)

ljharb:
@timhwang21 do you no longer have this use case? I'd still love to see this PR updated to match https://github.com/ljharb/qs/pull/226#issuecomment-657095653 so I can land it :-)

timhwang21:
I do not, and I unfortunately am not going to have time to work on this. I'm cleaning up my stale PRs because my dashboard is getting hard to manage.

ljharb:
It's really unfortunate that github's PR dashboard incentivizes people to close old PRs :-/ if this stays open, then someone else can take up the work very easily.

PR #229, opened 2017-10-04
Rename formats.default to formats.defaultFormat

such:
Having an attribute named 'default' confuses rollup when packaging.

See https://github.com/rollup/rollup/issues/1135 for instance.

### DIFF
--- a/lib/formats.js
+++ b/lib/formats.js
@@ -4,7 +4,7 @@ var replace = String.prototype.replace;
 var percentTwenties = /%20/g;
 
 module.exports = {
-    'default': 'RFC3986',
+    defaultFormat: 'RFC3986',
     formatters: {
         RFC1738: function (value) {
             return replace.call(value, percentTwenties, '+');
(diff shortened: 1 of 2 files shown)


dist/qs.js
```diff
@@ -80,36 +80,38 @@ var parseValues = function parseQueryStringValues(str, options) {
     return obj;
 };
 
-var parseObject = function parseObjectRecursive(chain, val, options) {
-    if (!chain.length) {
-        return val;
-    }
+var parseObject = function (chain, val, options) {
```
ljharb: regardless, all of these changes where you removed function names would have to be reverted.

such: I don't really understand where this change comes from... My changes in lib are restricted to the object of this PR...

ljharb: You might be rebased on a much older version of master?

ljharb: Oh, also, the dist file should never be updated except during a release :-) you can just revert this whole file.

btd:
iirc it is not an issue at all.

ljharb:
@btd you'll want to file a separate bug on rollup for that, then; a module bundler that doesn't understand a basic module format is broken.

I'm going to close this; tools should serve code, not the other way around.

PR #230, opened 2017-10-13
Change exports usage

btd:
Hello.

free `exports` usage (when it is not assignmnet for really exporting something) confusing rollup to wrap such modules in additional wrappers.

This PR remove `exports` for self referencing exporting symbols.

Thanks.

What do you think about replacing build with rollup (i mean dist/qs.js) - it will produce smaller file, and much readable/debuggable? (i can prepare PR)

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -34,7 +34,7 @@ var compactQueue = function compactQueue(queue) {
     return obj;
 };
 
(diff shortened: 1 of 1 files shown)


btd:
Yep, i understand what you mean. Thank you for merging.

Issue #219, opened 2017-07-04
feature request: support for arrayFormat: 'comma'


bryanlarsen:
from the documentation for go-querystring:

Slice and Array values default to encoding as multiple URL values of the same name. Including the "comma" option signals that the field should be encoded as a single comma-delimited value. Including the "space" option similarly encodes the value as a single space-delimited string. Including the "semicolon" option will encode the value as a semicolon-delimited string. Including the "brackets" option signals that the multiple URL values should have "[]" appended to the value name. "numbered" will append a number to the end of each incidence of the value name, example: name0=value0&name1=value1, etc.

The Kubernetes API appears to be using using the comma option, so it'd be nice to get support for that in qs.   Easy enough to work around, so it doesn't bother me enough to actually submit a pull request, but it's definitely a "would be nice".

ljharb:
This looks like it might be a duplicate of #112.

kmohrf:
I think I’d be nice if arrayFormat would just accept a function. than we can do whatever we want without without adding any more bloat for specialized array formats.

ljharb:
@kmohrf that's a very compelling and interesting suggestion. It seems like https://github.com/ljharb/qs/blob/master/lib/stringify.js#L91 could be the callsite, and https://github.com/ljharb/qs/blob/master/lib/stringify.js#L7-L15 are the three implementations for the currently supported formats, and https://github.com/ljharb/qs/blob/master/lib/stringify.js#L165 is where if `arrayFormat` is a function, it'd use that implementation instead of the alternatives.

kmohrf:
sounds good to me or was that an invitation to create a pull request? :)

michaelblash:
@ljharb I want to implement this feature but I see a flaw in the @kmohrf suggestion. If `arrayFormat` function accepts an array it should apparently return a string. This string is basically a number of pre-encoded values in a specific format, for example joined by commas. Therefore the user would have to implement his own encoding and even if he wants to utilize default encoder he would not be able to do this because the string should never be encoded any more.
For example, I need to encode an array `a: ['б', 'г']` in a comma separated format using a default encoder. I would do it like
```javascript
qs.stringify({ a: ['б', 'г'] }, {
  arrayFormat: inputArray => inputArray.map(
    a => encodeURIComponent(a) // (1)
  ).join(','),
  encoder: a => encodeURIComponent(a) // (2): duplication of (1)
})
```
[...]

PR #233, opened 2017-10-24
Allow strings to be split into arrays if desired

paul42:
more to scratch my own itch, but I was wondering if anyone else wanted to have the following functionality.
100% coverage, added entries to readme (love how that parses, btw) and tests for parse. tried to follow what I think was the coding conventions in the repo.

huge fan of the library, this is just a rough cut, can I get some feedback on it when you have a moment?

### DIFF
--- a/lib/parse.js
+++ b/lib/parse.js
@@ -83,7 +83,9 @@ var parseValues = function parseQueryStringValues(str, options) {
             val = interpretNumericEntities(val);
         }
 
-        if (val && options.comma && val.indexOf(',') > -1) {
+        if (options.valueParser !== null) {
+            val = options.valueParser(key, val, options);
+        } else if (val && options.comma && val.indexOf(',') > -1) {
             val = val.split(',');
         }
 
@@ -200,6 +202,10 @@ var normalizeParseOptions = function normalizeParseOptions(opts) {
     }
     var charset = typeof opts.charset === 'undefined' ? defaults.charset : opts.charset;
 
+    if (opts.valueParser !== null && opts.valueParser !== undefined && typeof opts.valueParser !== 'function') {
+        throw new TypeError('valueParser has to be a function.');
+    }
+
     return {
         allowDots: typeof opts.allowDots === 'undefined' ? defaults.allowDots : !!opts.allowDots,
         allowPrototypes: typeof opts.allowPrototypes === 'boolean' ? opts.allowPrototypes : defaults.allowPrototypes,
@@ -216,7 +222,8 @@ var normalizeParseOptions = function normalizeParseOptions(opts) {
         parameterLimit: typeof opts.parameterLimit === 'number' ? opts.parameterLimit : defaults.parameterLimit,
         parseArrays: opts.parseArrays !== false,
         plainObjects: typeof opts.plainObjects === 'boolean' ? opts.plainObjects : defaults.plainObjects,
-        strictNullHandling: typeof opts.strictNullHandling === 'boolean' ? opts.strictNullHandling : defaults.strictNullHandling
+        strictNullHandling: typeof opts.strictNullHandling === 'boolean' ? opts.strictNullHandling : defaults.strictNullHandling,
+        valueParser: valueParser
     };
 };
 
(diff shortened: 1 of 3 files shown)


lib/parse.js
```diff
@@ -144,9 +165,12 @@ module.exports = function (str, opts) {
 
     options.ignoreQueryPrefix = options.ignoreQueryPrefix === true;
     options.delimiter = typeof options.delimiter === 'string' || utils.isRegExp(options.delimiter) ? options.delimiter : defaults.delimiter;
+    options.arrayDelimiter = typeof options.arrayDelimiter === 'string' || utils.isRegExp(options.arrayDelimiter)
+        ? options.arrayDelimiter : defaults.arrayDelimiter;
     options.depth = typeof options.depth === 'number' ? options.depth : defaults.depth;
     options.arrayLimit = typeof options.arrayLimit === 'number' ? options.arrayLimit : defaults.arrayLimit;
     options.parseArrays = options.parseArrays !== false;
+    options.splitArrays = options.splitArrays !== false;
```
ljharb: in this case, `parseArrays` isn't the pattern to follow; this should ensure it's `=== true` instead. As is, this would be a breaking change.

ljharb (changes requested):
This is a different take on #219; it may be helpful to read that.

lib/parse.js
paul42: Hey thanks for taking the time to respond, I'm going to read through #219 and come back - I'm not dead set on any particular pattern, just happens to be my first pass, so please let me know if you see any potential issues or areas that desire more testing.

Thanks!

paul42: ah, so a function to split the value into arrays would remove the need for settings sprawl?

ljharb: yes, exactly

paul42:
is this more along the lines of what you were thinking?

ljharb:
Looks much closer! We should add the same function to stringification too.

paul42:
gotcha, let me take a crack at it

paul42:
@ljharb looking at this and 219 I think a solution could possibly be allowing an arrayFormatter parameter which would be a function, and if it's passed in it would supersede other options (like arrayPrefixes of brackets vs indices) and if it's not passed in or null, then the library operates as before - does that sound reasonable? I imagine it'll be a more advanced feature for those who want to fine tune their operation

paul42:
I'm also wondering if - to allow the function to act on all the items in the array, we should put the call to the function passed in up at https://github.com/ljharb/qs/blob/master/lib/stringify.js#L80 instead of inside the for loop for each key?

ljharb:
The function is called recursively, and it's important that it continue to do that - I think sticking with one key at a time is probably better.

paul42:
I'm having a bit of block seeing how to get from where it is to where I need to be on it - let me poke at it more, sorry for all the hand-holding that's required I'm still not super awesome at JS, so it's taking me a bit longer. I think the mental block I'm trying to over come is understanding how these lines work:
https://github.com/ljharb/qs/blob/master/lib/stringify.js#L91-L92
I'm not 100% on the syntax, so I don't follow what the second generateArrayPrefix does if nothing's passed in?

I think maybe I'll just go back and look up how to run the debugger and step through it a ton of times until I get a solid understanding of it.

ljharb:
In that case it just passes the *function itself*, since functions are first-class values in JS.

paul42:
so - in the case of object being an array, that value, 'generateArrayPrefix' gets passed parameters, does work, and returns - but in the case of not-an-array, the function is just passed in that order as a place holder so further recursive calls have it on hand to call? I think I'm grasping it, so I really can use the callsite you mention in 219. 💡  let me poke at it a bit more, thanks for the clarification.

dan-lee:
I'd really love to have this feature! Did you make some progress @paul42?

paul42:
hi @dan-lee sorry about the delay, but I'm having trouble wrapping my head around how to collapse the values into a single unit after it's reading the next key. if you want a really rough version you can try my branch, but it's not on npm and I haven't had a ton of time to work on this. I'll probably poke at it a bit more and then ask @ljharb for help again

michaelblash:
I'm really longing for this feature as well.

PR #234, opened 2017-11-08
Added ability to set no limits on arrays length (qs.parse arrayLimit option)

ps1dr3x:
Previous behaviour:

- Behaviour of arrayLimit = 0 option is the same of parseArrays = false
- The only way to parse arrays of any length is to set arrayLimit to a large value

New behaviour:

- By setting arrayLimit = 0, no length limits are considered


ps1dr3x:
My mistake, i didn't considered to set arrayLimit = Infinity

PR #236, opened 2017-11-18
Quote the sub-property for compatibility with closure-compiler

ChadKillingsworth:
2ed6ea4560cfb36eb6581b75bf148aa29b456210 broke compatibility with closure-compiler in ADVANCED mode.
Quoting this one key is all that is required to fix it.

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -168,7 +168,7 @@ var encode = function encode(str) {
 };
 
 var compact = function compact(value) {
-    var queue = [{ obj: { o: value }, prop: 'o' }];
+    var queue = [{ obj: { 'o': value }, prop: 'o' }]; // eslint-disable-line quote-props
     var refs = [];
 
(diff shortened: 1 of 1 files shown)


lib/utils.js
```diff
@@ -168,7 +168,7 @@ var encode = function encode(str) {
 };
 
 var compact = function compact(value) {
-    var queue = [{ obj: { o: value }, prop: 'o' }];
+    var queue = [{ obj: { 'o': value }, prop: 'o' }]; // eslint-disable-line quote-props
```
ljharb: If this breaks closure compiler in advanced mode, it's because closure compiler is broken in advanced mode. This is a valid unquoted property key, and it's not acceptable for a tool to require this kind of pollution to be pushed upstream.

I'd suggest using a non-broken bundling tool instead.

ChadKillingsworth: Got it - sorry to bother you. I'll go find a different library.

Issue #212, opened 2017-06-12
Nested arrays parsed incorrectly


oliversturm:
Hi,

I have this object:

```js
{
  filter: [
    [ 'int1', '=', 77 ],
    'and',
    [ 'int2', '=', 8 ]
  ]
}
```

Using jQuery `$.param()` (or jquery-param), this gives me the following query string:

```
filter%5B0%5D%5B%5D=int1&filter%5B0%5D%5B%5D=%3D&filter%5B0%5D%5B%5D=77&filter%5B%5D=and&filter%5B2%5D%5B%5D=int2&filter%5B2%5D%5B%5D=%3D&filter%5B2%5D%5B%5D=8

// decoded:
filter[0][]=int1&filter[0][]==&filter[0][]=77&filter[]=and&filter[2][]=int2&filter[2][]==&filter[2][]=8
```

Unfortunately, `qs.parse()` returns this instead of the original object:

```js
{
  filter: [ 
    [ 'int1', '=', '77', 'and' ], 
    [ 'int2', '=', '8' ] 
  ] 
}
```

The problem seems to be the part `filter[]=and` -- if I use `filter[1]=and` instead, the result is correct. However, the syntax makes sense to me the way it is, so I'm wondering if this is a bug in qs. Any ideas?

Thanks
Oli

ljharb:
What happens if you pass the `{ arrayFormat: 'indices' }` option to `qs.parse`? The default is `brackets`.

oliversturm:
@ljharb Unfortunately this doesn't seem to make any difference - I get the same result with or without any `arrayFormat`.

ljharb:
ahhhh yes.

If you use indices in an array, you have to use them everywhere - it's not like code where you can just stick a random `[]` in there and it acts like "push".

So, I think that either it all has to use `[]`, or, it all has to use indices.

oliversturm:
@ljharb Well, like I said it's not me doing this - jQuery `.param()` does it that way. And while I agree that it would look cleaner to have indices everywhere, the intention of the query string still seems clear and unambiguous to me. I can't judge whether this query string should be consider correct or not...

ljharb:
Any reason not to use qs.stringify?

oliversturm:
@ljharb Good question, it might be possible and I'll check it out. I guess it would still be good if qs understood the output from jQuery correctly ;)

oliversturm:
[...]

PR #242, opened 2017-12-26
Parse nested arrays correctly

jchook:
Re: #212 - Nested arrays parsed incorrectly

### DIFF
--- a/test/parse.js
+++ b/test/parse.js
@@ -237,6 +237,14 @@ test('parse()', function (t) {
         st.end();
     });
 
+    t.test('parses jquery-param strings', function (st) {
+        // readable = 'filter[0][]=int1&filter[0][]==&filter[0][]=77&filter[]=and&filter[2][]=int2&filter[2][]==&filter[2][]=8'
+        var encoded = 'filter%5B0%5D%5B%5D=int1&filter%5B0%5D%5B%5D=%3D&filter%5B0%5D%5B%5D=77&filter%5B%5D=and&filter%5B2%5D%5B%5D=int2&filter%5B2%5D%5B%5D=%3D&filter%5B2%5D%5B%5D=8';
+        var expected = { filter: [['int1', '=', '77'], 'and', ['int2', '=', '8']] };
+        st.deepEqual(qs.parse(encoded), expected);
+        st.end();
+    });
+
     t.test('continues parsing when no parent is found', function (st) {
         st.deepEqual(qs.parse('[]=&a=b'), { 0: '', a: 'b' });
         st.deepEqual(qs.parse('[]&a=b', { strictNullHandling: true }), { 0: null, a: 'b' });
--- a/test/utils.js
+++ b/test/utils.js
@@ -21,6 +21,30 @@ test('merge()', function (t) {
     var noOptionsNonObjectSource = utils.merge({ foo: 'baz' }, 'bar');
     t.deepEqual(noOptionsNonObjectSource, { foo: 'baz', bar: true });
 
+    t.test(
+        'avoids invoking array setters unnecessarily',
+        { skip: typeof Object.defineProperty !== 'function' },
+        function (st) {
+            var setCount = 0;
+            var getCount = 0;
+            var observed = [];
+            Object.defineProperty(observed, 0, {
+                get: function () {
+                    getCount += 1;
+                    return { bar: 'baz' };
+                },
+                set: function () { setCount += 1; }
+            });
+            utils.merge(observed, [null]);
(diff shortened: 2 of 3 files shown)


test/parse.js
```diff
@@ -236,6 +236,14 @@ test('parse()', function (t) {
         st.end();
     });
 
+    t.test('parses jquery-param strings', function (st) {
+        // readable = 'filter[0][]=int1&filter[0][]==&filter[0][]=77&filter[]=and&filter[2][]=int2&filter[2][]==&filter[2][]=8'
+        var encoded = 'filter%5B0%5D%5B%5D=int1&filter%5B0%5D%5B%5D=%3D&filter%5B0%5D%5B%5D=77&filter%5B%5D=and&filter%5B2%5D%5B%5D=int2&filter%5B2%5D%5B%5D=%3D&filter%5B2%5D%5B%5D=8';
```
ljharb: any chance we could programmatically construct `encoded` from `readable`, so the comment isn't needed?

test/utils.js
```diff
@@ -18,6 +18,15 @@ test('merge()', function (t) {
     var nestedArrays = utils.merge({ foo: ['baz'] }, { foo: ['bar', 'xyzzy'] });
     t.deepEqual(nestedArrays, { foo: ['baz', 'bar', 'xyzzy'] });
 
+    var setCount = 0;
+    var observed = [];
+    Object.defineProperty(observed, 0, {
```
ljharb: This should be inside a subtest that's skipped when `Object.defineProperty` is unavailable (for ES3 environments)

test/parse.js
jchook: We could. The OP mentioned this specific encoded string was produced by jQuery, so I intentionally hard-coded that exact input for the test to avoid any other (possibly changing) influences on what was being tested.

Using the readable string also produces the error without the fix, so we could just use that? Downside is we'd no longer be testing the exact `jQuery.param()` output. 

I could just remove the comment.

jchook: We could. The OP mentioned this specific encoded string was produced by jQuery, so I hard-coded that exact input for the test.

Using the readable string also produces the error without the fix, so we could just use that? Downside is we'd no longer be testing the exact jQuery.param() output.

I could just remove the comment.

test/utils.js
jchook: Let me know how that looks: c213b6bf15a22a9090a7a186401cbc0e75b0f23f

ljharb: Close! I'll comment on the commit.

test/utils.js
```diff
@@ -18,6 +18,22 @@ test('merge()', function (t) {
     var nestedArrays = utils.merge({ foo: ['baz'] }, { foo: ['bar', 'xyzzy'] });
     t.deepEqual(nestedArrays, { foo: ['baz', 'bar', 'xyzzy'] });
 
+    if (typeof Object.defineProperty === 'function') {
```
ljharb: instead, `t.test(msg, { skip: typeof Object.defineProperty !== 'function' }, function (st) {` should work

test/utils.js
```diff
@@ -18,6 +18,22 @@ test('merge()', function (t) {
...
+        t.test('avoids invoking array setters unnecessarily', function (st) {
+            var setCount = 0;
+            var observed = [];
+            Object.defineProperty(observed, 0, {
+                get: function () { return { bar: 'baz' }; },
```
ljharb: can we also keep track of `getCount` here?

ljharb:
@jchook is this still something you're interested in?

test/utils.js
jchook: > @jchook is this still something you're interested in?

Sure I'll take a look at your comments.

How should I handle getting the branch up-to-date? Rebase? New branch?

jchook:
Sure I'll take a look at your comments.

How should I fix the out-of-date branch? Rebase? New branch? No action needed?

test/utils.js
```diff
@@ -18,6 +18,30 @@ test('merge()', function (t) {
...
+                get: function () {
+                    getCount += 1;
+                    return { bar: 'baz' };
+                },
+                set: function () { setCount += 1; }
+            });
+            utils.merge(observed, [null]);
+            st.equal(setCount, 0);
+            st.equal(getCount, 2);
+            observed[0] = observed[0];
```
ljharb:
```suggestion
            observed[0] = observed[0]; // eslint-disable-line no-self-assign
```

jchook:
Ok I think it's good to go.

PR #274, opened 2018-08-27
ES Modules & CJS builds

43081j:
All major browsers now support ES modules natively, so it would be extremely useful to have qs be built in such a way it can be used with them.

Here's what I did:

* Rewrote the original sources to be ES modules (in `lib/`)
* Replaced browserify with rollup (which outputs a UMD bundle from the ESM sources to `dist/`)
* Added babel to transpile the ESM sources to CJS modules (in `lib/cjs`, added to gitignore)
* Added `module` to package.json so third-party tools can detect where the ESM sources live at
* Changed `main` to point at the transpiled CJS sources so node still works (`lib/cjs/index.js`)
* Changed all occurrences of `var` to `let` or `const` as appropriate

I left the tests as commonjs sources (they do not get transpiled) mostly so i could ensure the cjs output worked the same way as the original sources before this PR.

I tested this in a browser:

```html
<script type="module">
import * as qs from './node_modules/qs/lib/index.js';
qs.parse('a%5Bb%5D=c');
</script>
```

and it works fine.

What needs extra thought:

* Do the CJS sources really belong in `lib/cjs`? is there a better place? maybe have esm in `src/` and cjs in `lib/`?
* babel-cli won't work in old node versions afaik so CI will fail in those

---

Is this something you'd be interested in @ljharb? It would probably mean a major version bump (or would it? could get away with minor if all my export changes match up?).

If its not something you think you'll ever be interested in or prepared to merge, I'm happy to throw away the PR but it would be useful to have such native browser support.

Thanks!

cc @ljharb

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,0 +1,12 @@
+{
+    "presets": [
+        [
+            "@babel/preset-env",
+            {
+                "targets": {
+                    "node": true
+                }
+            }
+        ]
+    ]
+}
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,5 +1,9 @@
 {
     "root": true,
+    "parserOptions": {
+        "sourceType": "module",
+        "ecmaVersion": 6
+    },
 
     "extends": "@ljharb",
 
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@ node_modules
 
 # build output
 dist/*
+lib/cjs/
 
 # Only apps should have lockfiles
 yarn.lock
--- a/.npmignore
+++ b/.npmignore
@@ -1,4 +1,6 @@
 bower.json
 component.json
 .npmignore
+.babelrc
 .travis.yml
+rollup.config.js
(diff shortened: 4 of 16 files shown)


43081j:
Thanks for the quick reply on this. 

The addition of Babel to produce the common js sources means node still resolves to those rather than the ES sources (`main` still points to the CJS output). 

Rollup can be replaced with whatever bundler you prefer to be honest. I only changed it since it seemed simpler given the es sources. 

You already have a partial build due to the browser bundle so I figured it isn't a big jump to have another step. 

As for filing an issue first, that's true. But I don't mind, I wanted to see how straight forward (or not) it is to convert a node module so would have done this anyway, pushed or not :)

Again thanks for the response. Appreciate the feedback

ljharb:
I wouldn't want a bundler; bundlers are for apps.

Adding a babel build *at all* would be one PR to discuss (one I doubt would be worth it), and also producing ESM would be another - but i don't think there's any value in doing so until node supports ESM unflagged; browser support is irrelevant.

43081j:
You already have browserify, the rollup part of this pr doesn't matter much as browserify can still be used instead. 

So really it's just a question of if it's worth having a small build process for producing the commonjs sources. I can understand why you would want to keep the project "buildless" as it can be used as is, which is fair enough. 

You are right it would be much nicer if node supported esm without a flag.

Maybe then you would agree it's worth reinvestigation this once it does? By that point the build would fall away as the esm sources would work in both browser and node. 

It seems that the problem isn't with es modules but rather with the introduction of a build/complexity. And that only exists because of nodes lack of support.

ljharb:
Yes, I think it's worth investigating it once node ships ESM unflagged; it's not just about the complexity - i think it's dangerous and a mistake for anything to ship ESM until that point.

43081j:
that's fair enough. like i said originally i don't mind discarding the PR, its all a learning process.

hopefully node will support it more smoothly in future and without flags so we can just have one source for both node and browsers.

PR #275, opened 2018-08-30
allow arrayFormat to take a function

Primajin:
Allow for passing in a function to arrayFormat to help with #219

PR #276, opened 2018-09-09
add comma as an arrayFormat option

daggerjames:
It is a common feature request to parse array in 'comma' format, like #219 .

Simple allow we passed a function to arrayFormat, as #275 described, could not work, as 'comma' actually alter the left-side of equation, in a way that join the values of array rather than list them with well-structured key.  

In turn I create this new PR as an alternative way to solve it. However, it is still suffered some designed issue.

It would be hard to determine that how we represent complex array, such as `a=[{b: 'c'}]` in comma format. I simply use `JSON.stringify` to deal with inner objects, but I believe that there should be more common acknowledge to discuss about it.

Fortunately, most use-case of comma format would treat the values of array as a plain string, in turn there would be no problem with my approach. 

Further discussion would be welcomed.

### DIFF
--- a/lib/parse.js
+++ b/lib/parse.js
@@ -10,6 +10,7 @@ var defaults = {
     arrayLimit: 20,
     charset: 'utf-8',
     charsetSentinel: false,
+    comma: false,
     decoder: utils.decode,
     delimiter: '&',
     depth: 5,
@@ -81,6 +82,11 @@ var parseValues = function parseQueryStringValues(str, options) {
         if (val && options.interpretNumericEntities && charset === 'iso-8859-1') {
             val = interpretNumericEntities(val);
         }
+
+        if (val && options.comma && val.indexOf(',') > -1) {
+            val = val.split(',');
+        }
+
         if (has.call(obj, key)) {
             obj[key] = utils.combine(obj[key], val);
         } else {
@@ -200,6 +206,7 @@ var normalizeParseOptions = function normalizeParseOptions(opts) {
         arrayLimit: typeof opts.arrayLimit === 'number' ? opts.arrayLimit : defaults.arrayLimit,
         charset: charset,
         charsetSentinel: typeof opts.charsetSentinel === 'boolean' ? opts.charsetSentinel : defaults.charsetSentinel,
+        comma: typeof opts.comma === 'boolean' ? opts.comma : defaults.comma,
         decoder: typeof opts.decoder === 'function' ? opts.decoder : defaults.decoder,
         delimiter: typeof opts.delimiter === 'string' || utils.isRegExp(opts.delimiter) ? opts.delimiter : defaults.delimiter,
         depth: typeof opts.depth === 'number' ? opts.depth : defaults.depth,
--- a/lib/stringify.js
+++ b/lib/stringify.js
@@ -7,6 +7,7 @@ var arrayPrefixGenerators = {
     brackets: function brackets(prefix) { // eslint-disable-line func-name-matching
         return prefix + '[]';
     },
+    comma: 'comma',
     indices: function indices(prefix, key) { // eslint-disable-line func-name-matching
         return prefix + '[' + key + ']';
     },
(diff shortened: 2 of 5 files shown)


lib/stringify.js
```diff
@@ -7,6 +7,7 @@ var arrayPrefixGenerators = {
     brackets: function brackets(prefix) { // eslint-disable-line func-name-matching
         return prefix + '[]';
     },
+    comma: 'comma',
```
ljharb: why is this one not a function? that breaks with the established convention here.

lib/stringify.js
```diff
@@ -49,6 +50,10 @@ var stringify = function stringify( // eslint-disable-line func-name-matching
         obj = filter(prefix, obj);
     } else if (obj instanceof Date) {
         obj = serializeDate(obj);
+    } else if (generateArrayPrefix === 'comma' && Array.isArray(obj)) {
+        obj = obj.map(function (val) {
+            return typeof val === 'object' ? JSON.stringify(val) : val;
```
ljharb: why `JSON.stringify`? what if `val` also has an array, or also has a Date object, etc? I think this needs to recurse here.

daggerjames: this is why I wrote 'there should be more common acknowledge to discuss about it'

lib/stringify.js
daggerjames: Because 'comma' is differ to others in implementation, but would be more make sense to set in `arrayFormat`.

For example, what `brackets` or others do is try to iterate over each item in an array, with a well-documented key part in result, such as
``` 
[a,b,c].map(item => `${key(item)}=${item}`
```
, while `comma` is more like a joint stringify on value part
```
`${key(items)}=${[a,b,c].join(',')}`
```
in turn , `arrayPrefixGenerators['comma']` has no meaning as a function.

TejasQ:
Hello! @fabien0102 and I are engineers at @Contiamo and we use this feature in our library https://github.com/contiamo/restful-react. It would be quite helpful for us if this was merged.

Is there anything we can do to help ship this? 🙌

ljharb:
@TejasQ it's waiting on @daggerjames to update the PR to add a parse option.

daggerjames:
I'll check this night, sorry for delay

lib/stringify.js
daggerjames: I delete support for nested object.
`JSON.stringify` is not good enough, though I do not think further recurse would be appropriate. I don't see if it is a good idea to take all into account.

lib/parse.js
```diff
@@ -45,6 +46,14 @@ var parseValues = function parseQueryStringValues(str, options) {
     var skipIndex = -1; // Keep track of where the utf8 sentinel was found
     var i;
 
+    var updateKey = function updateObjKey(key, value) {
+        if (has.call(obj, key)) {
+            obj[key] = [].concat(obj[key]).concat(value);
```
ljharb: i believe we now have `utils.combine` to use here

lib/parse.js
```diff
@@ -45,6 +46,14 @@ var parseValues = function parseQueryStringValues(str, options) {
     var skipIndex = -1; // Keep track of where the utf8 sentinel was found
     var i;
 
+    var updateKey = function updateObjKey(key, value) {
```
ljharb: any reason not to take `obj` as an argument?

daggerjames: it's an inner function. 
if I take `obj` as a declared argument, it would be more reasonable to take the whole function out of `parseValues` function. 
I believe that it is more `close` to where it is actually called, but I can take it out.

lib/parse.js
daggerjames: thanks, I'll check

TejasQ:
Thanks for your great work, @daggerjames! You're a great help to us at @Contiamo! 💞

lib/parse.js
daggerjames: My bad, your original code phrase support set val with array. No need to strip anything out.  ：）

lib/parse.js
daggerjames: I check and I believe that it would be resolved during merging. 
I'll merge your main branch back to this PR as soon as there is no unresolved conversation left.

daggerjames:
hey, @ljharb , any update?

Issue #278, opened 2018-09-16
strictNullHandling not working when using a custom filter.


Neaox:
When using a custom filter then`strictNullHandling` does not work.

Neaox:
An explanation of the issue:

If you specify a custom filter in your stringify options the if case here (https://github.com/ljharb/qs/pull/279/files/ab07ce71f33704ecfac7298b055be8f481edf277#diff-dd5072864997b356b1727023f837c921L48) is run. Due the statement on line 52 (https://github.com/ljharb/qs/pull/279/files/ab07ce71f33704ecfac7298b055be8f481edf277#diff-dd5072864997b356b1727023f837c921L52) being an else if, instead of an if the null handling is not run, which then causes an error when attemping to get the keys of a null object.

PR #279, opened 2018-09-16
Allow `strictNullHandling` to work when using a custom filter

Neaox:
Fixes https://github.com/ljharb/qs/issues/278

### DIFF
--- a/lib/stringify.js
+++ b/lib/stringify.js
@@ -55,7 +55,9 @@ var stringify = function stringify( // eslint-disable-line func-name-matching
         obj = filter(prefix, obj);
     } else if (obj instanceof Date) {
         obj = serializeDate(obj);
-    } else if (obj === null) {
+    }
+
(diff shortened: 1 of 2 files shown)


Neaox:
Added a test as requested.

An explanation of the issue:

If you specify a custom filter in your stringify options the `if` case here (https://github.com/ljharb/qs/pull/279/files/ab07ce71f33704ecfac7298b055be8f481edf277#diff-dd5072864997b356b1727023f837c921L48) is run. Due the statement on line 52 (https://github.com/ljharb/qs/pull/279/files/ab07ce71f33704ecfac7298b055be8f481edf277#diff-dd5072864997b356b1727023f837c921L52) being an `else if`, instead of an `if` the null handling is not run, which then causes an error when attemping to get the keys of a null object.

By changing this `else if` to an `if` the issue is fixed.

PR #280, opened 2018-09-18
Convert build to to Webpack & Babel

joelove:


ljharb:
Seems like this was mistakenly meant to be https://github.com/joelove/qs/pull/1

PR #285, opened 2018-10-04
Improved checkJS

sandersn:
Moves most types back to JSDoc and fixes a few remaining type errors. Each commit has a discussion of what changed, and the commits are organised so that their diffs should individually be readable. I'm happy to discuss the hows and whys, since I want to understand what it's like to adopt checkJS for a JS expert.


benmccann:
There's a fork of this project with types that you may be interested in: https://www.npmjs.com/package/neoqs

PR #293, opened 2019-01-02
Introduces new `mapKeys` option

### LINKED ISSUE #291: Rename key while encoding?
Hey! Thanks for making this package, I'm finding it much more robust than the custom querystring constructor I wrote myself.

Here's my question: is there a way to rename a key while encoding its value? I'm wrestling with an API that expects the consumer to potentially provide `limit?: number` and `limit?: { // ... }` on some object, potentially at the same time. It's weird and frustrating, since I can't have two members with the same key on the query object. However, when expressed in the querystring `limit=20&limit[apps]=30`, it's technically valid. One way I'm getting around this is to name the object-typed `limit` key `limitField` to maintain symmetry with other parts of the query. In my custom stringify-er, I renamed the `limitField` key back to `limit` manually. Is it possible with the feature set that qs has today to perform that rename during processing?

aaronsky:
Fixes #291

This patch introduces a new option for the `stringify` and `parse`
functions to apply a transformation per key before encoding/after
decoding. 

This is still a WIP as I can't seem to get the test suite
passing locally. Any and all feedback would be appreciated!

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -11,11 +11,11 @@
         "indent": [2, 4],
         "max-lines-per-function": [2, { "max": 150 }],
         "max-params": [2, 14],
-        "max-statements": [2, 52],
+        "max-statements": [2, 53],
         "multiline-comment-style": 0,
         "no-continue": 1,
         "no-magic-numbers": 0,
         "no-restricted-syntax": [2, "BreakStatement", "DebuggerStatement", "ForInStatement", "LabeledStatement", "WithStatement"],
-        "operator-linebreak": [2, "before"],
+        "operator-linebreak": [2, "before"]
     }
 }
--- a/README.md
+++ b/README.md
@@ -202,6 +202,15 @@ assert.deepEqual(detectedAsIso8859_1, { a: '☺' });
 It also works when the charset has been detected in `charsetSentinel`
 mode.
 
+In addition, you can transform parsed keys after decoding with `mapKeys`:
+
+```javascript
+assert.deepEqual(
+    qs.parse('aField=7', { mapKeys: function (k) { return k === 'aField' ? 'a' : k; } }),
+    { a: 7 }
+);
+```
+
 ### Parsing Arrays
 
 **qs** can also parse arrays using a similar `[]` notation:
@@ -407,6 +416,15 @@ assert.equal(
     qs.stringify({ a: date }, { serializeDate: function (d) { return d.getTime(); } }),
     'a=7'
 );
+```
+
+In addition, you can transform keys entirely without writing a custom encoder using `mapKeys`:
+
+```javascript
+assert.equal(
+    qs.stringify({ a__rename: 7 }, { mapKeys: function(k) { return k === 'a__rename' ? 'a' : k; } }),
+    'a=7'
+)
 ```
 
 You may use the `sort` option to affect the order of parameter keys:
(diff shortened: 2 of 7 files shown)


.eslintrc
```diff
@@ -11,11 +11,11 @@
...
         "max-lines-per-function": [2, { "max": 150 }],
         "max-params": [2, 14],
-        "max-statements": [2, 52],
+        "max-statements": [2, 53],
         "multiline-comment-style": 0,
         "no-continue": 1,
         "no-magic-numbers": 0,
         "no-restricted-syntax": [2, "BreakStatement", "DebuggerStatement", "ForInStatement", "LabeledStatement", "WithStatement"],
-        "operator-linebreak": [2, "before"],
+        "operator-linebreak": [2, "before"]
```
ljharb: trailing commas are fine in eslintrc files, fwiw
```suggestion
        "operator-linebreak": [2, "before"],
```

README.md
```diff
@@ -409,6 +418,15 @@ assert.equal(
 );
 ```
 
+In addition, you can transform keys entirely without writing a custom encoder using `mapKeys`:
```
ljharb: how does this interact with a custom encoder? are the options mutually exclusive, or is there a specific order they run in? (same question re a custom decoder in parse)

dist/qs.js
```diff
@@ -578,7 +578,7 @@ var merge = function merge(target, source, options) {
     if (Array.isArray(target) && Array.isArray(source)) {
         source.forEach(function (item, i) {
             if (has.call(target, i)) {
-                if (target[i] && typeof target[i] === 'object') {
+                if (target[i] && typeof target[i] === 'object' && item && typeof item === 'object') {
```
ljharb: please revert this; the dist file should only be updated at release time.
```suggestion
                if (target[i] && typeof target[i] === 'object') {
```

lib/parse.js
```diff
@@ -78,6 +81,8 @@ var parseValues = function parseQueryStringValues(str, options) {
             val = options.decoder(part.slice(pos + 1), defaults.decoder, charset);
         }
 
+        key = options.mapKeys(key);
```
ljharb: this should be extracted out so that the `mapKeys` function does not get `options` as its receiver (`this` value)

lib/stringify.js
```diff
@@ -87,9 +92,9 @@ var stringify = function stringify( // eslint-disable-line func-name-matching
 
     var objKeys;
     if (Array.isArray(filter)) {
-        objKeys = filter;
+        objKeys = filter.map(mapKeys);
```
ljharb: this will cause three arguments to be passed to `mapKeys`, perhaps we want `function (k) { return mapKeys(k); }` to limit the observable API?

lib/stringify.js
```diff
@@ -87,9 +92,9 @@ var stringify = function stringify( // eslint-disable-line func-name-matching
...
     } else {
-        var keys = Object.keys(obj);
+        var keys = Object.keys(obj).map(mapKeys);
```
ljharb: same here

tamdao:
Any progress with this PR?

ljharb:
@tamdao nope.

@aaronsky are you still interested in completing this PR?

ljharb:
@aaronsky so you’re no longer interested?

PR #307, opened 2019-04-12
Skipping blank values at stringify

douglira:


dist/qs.js
```diff
@@ -308,6 +309,7 @@ var stringify = function stringify( // eslint-disable-line func-name-matching
     prefix,
     generateArrayPrefix,
     strictNullHandling,
+    skipBlanks,
```
ljharb: All changes made in this autogenerated file are lost; you'll need to change the files in `lib` instead.

ljharb (changes requested):
Hmm - "blank" isn't really a universal definition. Perhaps this would be better as a generic `skip` option? It would take a function, be passed key and value, and return `true` to skip. That way it could also subsume the `skipNulls` option (it would throw if both were passed), and then you could define your own definition for what to skip.

We'd want to add the same option to both parse and stringify.

douglira:
It makes sense and more suitable. I will give it a try asap

ljharb:
@douglira have you had any time to update this PR?

PR #312, opened 2019-05-27
[New] `parse`: add `allowSparse` option for collapsing arrays with missing indices

### LINKED ISSUE #181: Option to not collapse arrays with missing indices.
The following test:

    $ node -pe 'require("qs").parse("option[1]=1&text[0]=&text[1]=&text[2]=")'

Produces:

    { option: [ '1' ], text: [ '', '', '' ] }

The `option` array is collapsed and the missing index 0 is removed. According to the qs documentation, this behavior is intentional. However, it causes problems when handling arrays of checkboxes on a form with e.g. body-parser.

Consider:

index.js:

    var express = require("express");
    var site = express();
    
    site.use(require("body-parser").urlencoded({extended:true}));
    
    site.get("/", function (req, res) {
    	res.sendFile(__dirname + "/test.html");
    });
    
    site.post("/form", function (req, res) {
    	res.json(req.body);
    });
    
    site.listen(8081);

test.html:

    <html>
    	<body>
    		<form method="post" action="/form">
    			<input type="checkbox" name="option[0]" value="1">
    			<input type="checkbox" name="option[1]" value="1">
    			<input type="checkbox" name="option[2]" value="1"><br>
    			<input type="text" name="text[0]"><br>
    			<input type="text" name="text[1]"><br>
    			<input type="text" name="text[2]"><br>
    			<button type="submit">Submit</button>
    		</form>
    	</body>
    </html>

In that example, if I were to check only `option[1]`, I verify that the index is set correctly by inspecting the request in Chrome, body contents are:

    option[1]:1
    text[0]:
    text[1]:
    text[2]:

Yet qs collapses the `option` array and body-parser ultimately produces the following in `req.body`:

    {"option":["1"],"text":["","",""]}

Similarly, if I were to check `option[0]` and `option[2]`, the request body would have:

    option[0]:1
    option[2]:1
    text[0]:
    text[1]:
    text[2]:

But it would be parsed to:

    {"option":["1","1"],"text":["","",""]}

All information about which checkbox was checked is lost.

The only way, then, to parse checkbox arrays on forms is workarounds like the clever but hacky one at http://stackoverflow.com/a/41093135, which basically involves hidden inputs with the same name as the checkboxes, then using `Array.isArray()` to determine if a result is true or false.

I had originally opened this issue at https://github.com/expressjs/body-parser/issues/211 before I knew it was qs.

To that end, I'd like an option to *not* collapse arrays, such that the previous examples yield things like:

    {"option":[null,"1"]}

Or:

    {"option":["1",null,"1"]}

This would make parsing arrays of checkboxes on forms possible, which it currently is not without trickery.

1cheese:
Fixes #181

### DIFF
--- a/lib/parse.js
+++ b/lib/parse.js
@@ -8,6 +8,7 @@ var isArray = Array.isArray;
 var defaults = {
     allowDots: false,
     allowPrototypes: false,
+    allowSparse: false,
     arrayLimit: 20,
     charset: 'utf-8',
     charsetSentinel: false,
@@ -217,6 +218,7 @@ var normalizeParseOptions = function normalizeParseOptions(opts) {
     return {
         allowDots: typeof opts.allowDots === 'undefined' ? defaults.allowDots : !!opts.allowDots,
         allowPrototypes: typeof opts.allowPrototypes === 'boolean' ? opts.allowPrototypes : defaults.allowPrototypes,
+        allowSparse: typeof opts.allowSparse === 'boolean' ? opts.allowSparse : defaults.allowSparse,
         arrayLimit: typeof opts.arrayLimit === 'number' ? opts.arrayLimit : defaults.arrayLimit,
         charset: charset,
         charsetSentinel: typeof opts.charsetSentinel === 'boolean' ? opts.charsetSentinel : defaults.charsetSentinel,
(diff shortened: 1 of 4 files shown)


lib/parse.js
```diff
@@ -10,6 +10,7 @@ var defaults = {
     arrayLimit: 20,
     charset: 'utf-8',
     charsetSentinel: false,
+    collapseArrayIndices: true,
```
ljharb: it's better if boolean options default to `false`, so that absence and false are the same. can we rename this option so it can default to `false`?

1cheese: Renamed this option to `allowSparse`.

ljharb (approved):
I've rebased this, and added stringify tests to match the parse tests that you added (thanks).

I continue to think producing sparse arrays is a bad idea, but since stringify accepts them, it seems useful to enable a round trip.

PR #325, opened 2019-07-24
change default arrayFormat to repeat

cloud411:
This change is required so that the keys are not the default.

PS.
before commit
qs.stringify({ id: ['a', 'b'] }); => id%5B0%5D=a&id%5B1%5D=b (id[0]=a&id[1]=b)

after commit
qs.stringify({ id: ['a', 'b'] }); => id=a&id=b

### DIFF
--- a/lib/stringify.js
+++ b/lib/stringify.js
@@ -226,7 +226,7 @@ module.exports = function (object, opts) {
     } else if (opts && 'indices' in opts) {
         arrayFormat = opts.indices ? 'indices' : 'repeat';
     } else {
-        arrayFormat = 'indices';
+        arrayFormat = 'repeat';
     }
 
     var generateArrayPrefix = arrayPrefixGenerators[arrayFormat];


cloud411:
Your package uses the famous **request**.
And when you add query parameters to options, eg:
`
{
...
qs: { id: [1, 2] }
}
`
Then the request is sent with keys that you do not expect at the other end

cloud411:
I believe that by default the query parameters must be sent as they were passed

cloud411:
Also, the behavior of your package is different from the behavior of packages **querystring**
That de facto is the standard in the world **Node.js**

cloud411:
https://ibb.co/pZYYGqT

ljharb:
If they're passed as an array, they should be sent in such a way as to indicate that they are - using brackets.

It doesn't really matter what `querystring` does; qs isn't a polyfill for that. `qs` exists because `querystring` wasn't good enough.

cloud411:
OK. I understood you

PR #329, opened 2019-09-06
arrayFormat for parse

dreyks:
added basic tests. without arrayFormat parse's behavior does not change

don't think we want to allow specifying several formats simultaneously?

### DIFF
--- a/test/parse.js
+++ b/test/parse.js
@@ -34,31 +34,31 @@ test('parse()', function (t) {
 
     t.test('arrayFormat: brackets allows only explicit arrays', function (st) {
         st.deepEqual(qs.parse('a[]=b&a[]=c', { arrayFormat: 'brackets' }), { a: ['b', 'c'] });
-        st.deepEqual(qs.parse('a[0]=b&a[1]=c', { arrayFormat: 'brackets' }), { a: ['b', 'c'] });
+        st.deepEqual(qs.parse('a[0]=b&a[1]=c', { arrayFormat: 'brackets' }), { 'a[0]': 'b', 'a[1]': 'c' });
         st.deepEqual(qs.parse('a=b,c', { arrayFormat: 'brackets' }), { a: 'b,c' });
-        st.deepEqual(qs.parse('a=b&a=c', { arrayFormat: 'brackets' }), { a: ['b', 'c'] });
+        st.deepEqual(qs.parse('a=b&a=c', { arrayFormat: 'brackets' }), { a: 'c' });
         st.end();
     });
 
     t.test('arrayFormat: indices allows only indexed arrays', function (st) {
-        st.deepEqual(qs.parse('a[]=b&a[]=c', { arrayFormat: 'indices' }), { a: ['b', 'c'] });
+        st.deepEqual(qs.parse('a[]=b&a[]=c', { arrayFormat: 'indices' }), { 'a[]': 'c' });
         st.deepEqual(qs.parse('a[0]=b&a[1]=c', { arrayFormat: 'indices' }), { a: ['b', 'c'] });
         st.deepEqual(qs.parse('a=b,c', { arrayFormat: 'indices' }), { a: 'b,c' });
-        st.deepEqual(qs.parse('a=b&a=c', { arrayFormat: 'indices' }), { a: ['b', 'c'] });
+        st.deepEqual(qs.parse('a=b&a=c', { arrayFormat: 'indices' }), { a: 'c' });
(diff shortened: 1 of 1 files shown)


ljharb:
Are these tests expected to pass as-is, or fail?

dreyks:
they’re expected to fail. i just wanted to be sure that this is the behavior we want. if you’re ok with it - i’ll proceed to the implementation

ljharb:
@dreyks i think it would be helpful to have two commits - one with passing tests, and another with the tests that fail. It makes it much easier for me to see what's breaking or not, and what requires a new option or not.

dreyks:
oh. gotcha. ok will do

dreyks:
done

test/parse.js
```diff
@@ -34,17 +34,31 @@ test('parse()', function (t) {
 
     t.test('arrayFormat: brackets allows only explicit arrays', function (st) {
         st.deepEqual(qs.parse('a[]=b&a[]=c', { arrayFormat: 'brackets' }), { a: ['b', 'c'] });
+        st.deepEqual(qs.parse('a[0]=b&a[1]=c', { arrayFormat: 'brackets' }), { 'a[0]': 'b', 'a[1]': 'c' });
```
ljharb: i see how this seems correct, because it's currently set to brackets but is parsing indices.

However, the current behavior matches what PHP, Rails, and express all do, so I'm not sure it should change.

test/parse.js
```diff
@@ -34,31 +34,31 @@ test('parse()', function (t) {
 
     t.test('arrayFormat: brackets allows only explicit arrays', function (st) {
         st.deepEqual(qs.parse('a[]=b&a[]=c', { arrayFormat: 'brackets' }), { a: ['b', 'c'] });
-        st.deepEqual(qs.parse('a[0]=b&a[1]=c', { arrayFormat: 'brackets' }), { a: ['b', 'c'] });
+        st.deepEqual(qs.parse('a[0]=b&a[1]=c', { arrayFormat: 'brackets' }), { 'a[0]': 'b', 'a[1]': 'c' });
         st.deepEqual(qs.parse('a=b,c', { arrayFormat: 'brackets' }), { a: 'b,c' });
-        st.deepEqual(qs.parse('a=b&a=c', { arrayFormat: 'brackets' }), { a: ['b', 'c'] });
+        st.deepEqual(qs.parse('a=b&a=c', { arrayFormat: 'brackets' }), { a: 'c' });
```
ljharb: i see how this seems correct, because it's currently set to brackets but is parsing as repeat.

However, the current behavior matches what PHP, Rails, and express all do, so I'm not sure it should change.

dreyks: Rails and PHP actually parse it as overwriting

this is the reason I've started all this: in my rails app I've switched from "chaos and anarchy " to qs and got one spec failure where the suite was appending params and with qs that lead to an array

test/parse.js
dreyks: yeah, i'm not very sure on this one too, though you have to explicitly state you want this behavior so maybe it's ok

alternatively making allowing `arrayFormat` to be an array itself could solve it

```js
t.deepEqual(qs.parse('a[0]=b&a[1]=c', { arrayFormat: ['brackets', 'indices'] }), { 'a': ['b',  'c'] });
t.deepEqual(qs.parse('a[]=b&a[]=c', { arrayFormat: ['brackets', 'indices'] }), { 'a': ['b',  'c'] });
```

test/parse.js
ljharb: hmm, thanks for correcting me then. i'll think on this one.

AlanKnightly:
How is this feature going?🧐

ljharb:
@AlanKnightly you can see that for yourself, without creating notification noise on this PR and the linked issue, by merely reading the comments above.

ljharb:
@dreyks ok, so, here's how i think this needs to work:

 - `arrayFormat` must take all the options stringify does - brackets, indices, repeat, comma - and produce an object that, when re-stringified with the same format, produces the original parse input (ie, roundtripping)
 - The current behavior *must* be described by one of the `arrayFormat` options. if none of the 4 formats above match, we may need to make a 5th one.
 
One important goal is that the default behavior of Rails, PHP, and express should all be expressible with an option - ideally the same one, and I suspect that's "brackets".

It'd be great to get this revived :-)

dreyks:
oof, lemme recollect what's going on here, it's been 4 years and not the easiest ones :D

Issue #290, opened 2018-12-25
add options of ignoring uppercase and lowercase


silaLi:
When I use it, I can't be sure whether it is uppercase or lowercase. At this point I need to write a lot of parameters

ljharb:
Can you explain the use case a bit more? What's the current client and server behavior?

Neaox:
I am needing similar functionality. In my case it refers to `parse`.

Given the following string:

`?propA=value1&propB=value2&PropA=value3`

I would like to be able to get the following:

```
{
  "propa": ["value1", "value3"],
  "propb": "value2"
}
```

I was hoping to do this using a custom decoder function, however, the decoder function is used to decode both the key and value and there is no way inside the decode function to know if you are dealing with a key or a value.

ljharb:
@Neaox it would be reasonable to pass to the decoder the string "key" or "value", to indicate which one it should be - i'd be open to a PR to add that to both parse-decode and stringify-encode.

Neaox:
I've created a PR to introduce this functionality.

Neaox:
When can we expect the next version to be released?

ljharb:
v2.9.0 is released.

PR #333, opened 2019-09-12
Provide type argument to encoder/decoder function used by parse/stringify

Neaox:
Addresses: https://github.com/ljharb/qs/issues/290


Neaox:
Updated the README as requested.

Issue #328, opened 2019-08-27
parse: comma: true assumes string values limits use of decoder


Nettsentrisk:
If you want to decode numeric values as numbers, through the use of the decoder option, you could pass in the following decoder function:

```javascript
decoder(value) {
            if (/^(\d+|\d*\.\d+)$/.test(value)) {
              return parseFloat(value);
            }
            return value;
          }
```
However, when combined with comma: true, you will get an indexOf error due to the fact that the values have been decoded to numbers before being parsed by the code splitting them up via commas, thus breaking this parsing since it assumes all values are strings.

This could associated with splitting on commas should be run before decoding takes place, or there needs to be a check to see whether the value being checked for commas is actually a string.

PR #334, opened 2019-10-04
[fix] comma option enabled value parsing to use strings only

Om4ar:
in case of comma enabled .. parsing should check for strings only when using indexOf(',')

fixes: #328

### DIFF
--- a/lib/parse.js
+++ b/lib/parse.js
@@ -83,7 +83,7 @@ var parseValues = function parseQueryStringValues(str, options) {
             val = interpretNumericEntities(val);
         }
 
-        if (val && options.comma && val.indexOf(',') > -1) {
+        if (val && typeof val === 'string' && options.comma && val.indexOf(',') > -1) {
             val = val.split(',');
         }
 
--- a/test/parse.js
+++ b/test/parse.js
@@ -400,6 +400,20 @@ test('parse()', function (t) {
         st.end();
     });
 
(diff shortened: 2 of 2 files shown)


dist/qs.js
```diff
@@ -125,7 +125,7 @@ var parseValues = function parseQueryStringValues(str, options) {
             val = interpretNumericEntities(val);
         }
 
-        if (val && options.comma && val.indexOf(',') > -1) {
+        if (val && typeof val === 'string' && options.comma && val.indexOf(',') > -1) {
```
ljharb: please revert this; this file is only updated at release time.

lib/parse.js
```diff
@@ -83,7 +83,7 @@ var parseValues = function parseQueryStringValues(str, options) {
             val = interpretNumericEntities(val);
         }
 
-        if (val && options.comma && val.indexOf(',') > -1) {
+        if (val && typeof val === 'string' && options.comma && val.indexOf(',') > -1) {
```
ljharb: can you please add a regression test?

main.js
```diff
@@ -0,0 +1,3 @@
+const { parse } = require("./lib");
```
ljharb: by accident, i assume?

Om4ar: yes just removed it

ljharb:
The test you added passes without your fix - can you tweak the test so it fails without the fix?

Om4ar:
@ljharb  thank you 
i learned a lot from this

PR #335, opened 2019-10-04
[fix] parsing a field that holds array of arrays

### LINKED ISSUE #327: comma and bracket combined?
```
parse("foo[]=1,2,3&foo[]=4,5,6", {comma: true})
/*
expected { foo: [ [ 1, 2, 3 ], [4, 5, 6] ] }
received { foo: [ 1, 2, 3, 4, 5, 6 ] }

*/
```

uncertain if this is intentional or not

Om4ar:
fixes #327

### DIFF
--- a/lib/parse.js
+++ b/lib/parse.js
@@ -3,6 +3,7 @@
 var utils = require('./utils');
 
 var has = Object.prototype.hasOwnProperty;
+var isArray = Array.isArray;
 
 var defaults = {
     allowDots: false,
@@ -85,6 +86,10 @@ var parseValues = function parseQueryStringValues(str, options) {
 
         if (val && typeof val === 'string' && options.comma && val.indexOf(',') > -1) {
             val = val.split(',');
+        }
+
+        if (part.indexOf('[]=') > -1) {
+            val = isArray(val) ? [val] : val;
         }
 
         if (has.call(obj, key)) {
--- a/test/parse.js
+++ b/test/parse.js
@@ -414,6 +414,15 @@ test('parse()', function (t) {
         st.end();
     });
 
+    t.test('parses brackets holds array of arrays when having two parts of strings with comma as array divider', function (st) {
+        st.deepEqual(qs.parse('foo[]=1,2,3&foo[]=4,5,6', { comma: true }), { foo: [['1', '2', '3'], ['4', '5', '6']] });
+        st.deepEqual(qs.parse('foo[]=1,2,3&foo[]=', { comma: true }), { foo: [['1', '2', '3'], ''] });
+        st.deepEqual(qs.parse('foo[]=1,2,3&foo[]=,', { comma: true }), { foo: [['1', '2', '3'], ['', '']] });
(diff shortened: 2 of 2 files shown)


lib/parse.js
```diff
@@ -87,6 +87,10 @@ var parseValues = function parseQueryStringValues(str, options) {
             val = val.split(',');
         }
 
+        if(part.indexOf('[]=') > -1){
```
ljharb:
``` suggestion
        if (part.indexOf('[]=') > -1) {
```
what happens with `foo[][]=bar`, or just `[]=a`?

lib/parse.js
```diff
@@ -87,6 +87,10 @@ var parseValues = function parseQueryStringValues(str, options) {
...
+          val = Array.isArray(val) ? [val] : val
```
ljharb: qs supports ES3, so we can't rely on `Array.isArray` being there

Om4ar: @ljharb 
replaced in utils
`var isArray = Array.isArray;` this is being used in the utils file
with
`var isArray = Object.prototype.toString.call(obj) === '[object Array]';`

lib/parse.js
```diff
@@ -87,6 +87,10 @@ var parseValues = function parseQueryStringValues(str, options) {
             val = val.split(',');
         }
 
+        if (part.indexOf('[]=') > -1) {
+            val = utils.isArray(val) ? [val] : val;
```
ljharb: a perhaps simpler alternative here:
```suggestion
            val = [].concat([val]);
```

Om4ar: @ljharb 

`val = [].concat([val]);`

will fail test `allows disabling array parsing` in the following case:

```
        var emptyBrackets = qs.parse('a[]=b', { parseArrays: false });
        st.deepEqual(emptyBrackets, { a: { 0: 'b' } });
```
expected: |-
      { a: { 0: 'b' } }
    actual: |-
      { a: { 0: [ 'b' ] } }

lib/parse.js
Om4ar: @ljharb 

`foo[][]=bar` will be parsed to ```{ foo: ['bar'] }```

`[]=a` will be parsed to  ```{ 0: 'a' }```

my code doesn't affect these results, it will be the same result

lib/parse.js
ljharb: ah, right, `[].concat([val])` is actually just `[val]`, so that's not the same.

lib/utils.js
```diff
@@ -1,7 +1,10 @@
 'use strict';
 
 var has = Object.prototype.hasOwnProperty;
-var isArray = Array.isArray;
```
ljharb: whoops, turns out we already depended on this being available ¯\\\_(ツ)_/¯

Om4ar: @ljharb  yes it is being used in stringify.js file too =D

Om4ar: @ljharb 
should i revert my changes on utils file and make another pr to replace Array.isArray by the polyfill ?

ljharb: nah for now just use Array.isArray the same way it's used in stringify.js.

Om4ar: okay, will revert my is changes for using this pollyfill

PR #336, opened 2019-10-18
[fix] parsing comma delimited str with percent encoded comma

### LINKED ISSUE #311: Comma-delimited array parsing fails when individual items contain commas
Comma-delimited array parsing fails when running something like this:

```javascript
Qs.parse('things=a%2C%20b,c%2C%20d', {
  comma: true
});
```

Expected: `{ things: ['a, b', 'c, d'] }`
Actual: `{ things: ['a', ' b', 'c', ' d'] }`

Per my understanding of RFC 3986, a percent-encoded comma should not be treated the same as a comma literal. Comma literal is a reserved character used for delimiting parameters and parameter values, but a percent-encoded comma is not.

My use case needs to deal with arbitrary user input within values, potentially including commas.

I guess this can be fixed with a custom `decoder` function, but it seems like something that should work by default when `comma` is `true`.

As my use case is otherwise fairly simple and I don't need lots of custom features, I figured I'd just roll [my own query string parser/serializer](https://gist.github.com/lionel-rowe/abdf3a128e0c22a8f5123e9bb4007891). I'm linking to this strictly for reference and wouldn't recommend anyone to use it in production in its current state - it probably diverges from the RFC in a million other ways and fails for lots of edge cases that aren't relevant to my specific use case.

Related: #237 (but that deals with `stringify`, whereas this is about `parse`).

Om4ar:
changes on the default decoder to correctly parse comma delimited strings while having percent encoded comma

Fixes #311.

### DIFF
--- a/lib/parse.js
+++ b/lib/parse.js
@@ -85,7 +85,15 @@ var parseValues = function parseQueryStringValues(str, options) {
             val = options.strictNullHandling ? null : '';
         } else {
             key = options.decoder(part.slice(0, pos), defaults.decoder, charset, 'key');
-            val = options.decoder(part.slice(pos + 1), defaults.decoder, charset, 'value');
+            var encodedVal = part.slice(pos + 1);
+            if (options.comma && encodedVal.indexOf(',') !== -1) {
+                val = encodedVal.split(',')
+                    .map(function (encodedFragment) {
+                        return options.decoder(encodedFragment, defaults.decoder, charset, 'value');
+                    });
+            } else {
+                val = options.decoder(encodedVal, defaults.decoder, charset, 'value');
+            }
         }
 
         if (val && options.interpretNumericEntities && charset === 'iso-8859-1') {
--- a/test/parse.js
+++ b/test/parse.js
@@ -429,6 +429,14 @@ test('parse()', function (t) {
         st.end();
     });
 
+    t.test('parses comma delimited array while having percent-encoded comma treated as normal text', function (st) {
+        st.deepEqual(qs.parse('foo=a%2Cb', { comma: true }), { foo: ['a', 'b'] });
+        st.deepEqual(qs.parse('foo=a%2C%20b,d', { comma: true }), { foo: ['a, b', 'd'] });
+        st.deepEqual(qs.parse('foo=a%2C%20b,c%2C%20d', { comma: true }), { foo: ['a, b', 'c, d'] });
+
+        st.end();
+    });
+
     t.test('parses an object in dot notation', function (st) {
         var input = {
             'user.name': { 'pop[bob]': 3 },


lib/parse.js
```diff
@@ -77,17 +77,13 @@ var parseValues = function parseQueryStringValues(str, options) {
...
-            val = options.decoder(part.slice(pos + 1), defaults.decoder, charset, 'value');
+            val = options.decoder(part.slice(pos + 1), defaults.decoder, charset, 'value', options.comma);
         }
 
         if (val && options.interpretNumericEntities && charset === 'iso-8859-1') {
             val = interpretNumericEntities(val);
         }
 
-        if (val && typeof val === 'string' && options.comma && val.indexOf(',') > -1) {
-            val = val.split(',');
```
ljharb: Removing this is a breaking change; it forces anyone using a custom decoder to add this logic where they previously didn't require it.

Om4ar: i agree, i will look into it more to see what can be done

quentinus95:
You can pull a non-breaking version here: https://github.com/quentinus95/qs/tree/comma-perecent-parsing

ljharb:
@quentinus95 pulled in and rebased, but now the added test is failing.

quentinus95:
@ljharb It is due to the rebase.

Can you remove call on line https://github.com/ljharb/qs/pull/336/files#diff-9f97a9c36fbbfb0d19068d6fae3bb84eR103?

Thanks

ljharb:
@quentinus95 that was added in #359, and is a pretty important bugfix - so this PR may have to be adapted to work with it.

quentinus95:
@ljharb thanks, I've just had a look over it, and it would take too long for me to support the fix (especially because #359 has duplicated some of the logic). I've started to do some refactoring but won't have time to achieve it.

If you have some hints on how to fix this without rewriting 20% of the parser, I'll be happy to write the implementation.

ljharb:
Looks like just the test needs to be updated.

test/parse.js
```diff
@@ -429,6 +429,14 @@ test('parse()', function (t) {
         st.end();
     });
 
+    t.test('parses comma delimited array while having percent-encoded comma treated as normal text', function (st) {
+        st.deepEqual(qs.parse('foo=a%2Cb', { comma: true }), { foo: ['a', 'b'] });
```
quentinus95: @ljharb this test is wrong, it should return "a,b". It's the main point of the pull request and the related issue.

ljharb: ah, i understand. my bad.

ljharb: Filed https://github.com/ljharb/qs/pull/361 to proceed with a proper fix.

PR #338, opened 2019-10-18
[update] stringify with comma:true to decode comma for an array as a reserved char per RFC3986

### LINKED ISSUE #337: Stringify with arrayFormat: comma, wrongfully encode comma in return
wrong result with stringifying 

```qs.stringify({ a: ['b', 'c'] }, { arrayFormat: 'comma' })```
return ```a=b%2Cc```

when it should return ```a=b,c```

because parsing ```,```  as reserved character per RFC 3986 should split the array if we used a parser with```comma: true``` option 
and ```%2C```  (percent encoded comma) should be treated normally as text character
 
related to : #336 pr

### PR DESCRIPTION
**Pull request by Om4ar**
pr to correctly stringify a array with formatArray as a comma

keeping the comma character decoded to separate items for this character a reserved one per RFC3986

fixes: #337

### DIFF
--- a/lib/stringify.js
+++ b/lib/stringify.js
@@ -100,6 +100,21 @@ var stringify = function stringify(
         return [formatter(prefix) + '=' + formatter(String(obj))];
     }
 
+    if (generateArrayPrefix === 'comma' && isArray(obj)) {
+        if (formatter !== formats.formatters['RFC3986']) {
+            obj = obj.join(',');
+        } else {
+            if (encoder) {
+                var commaKey = encodeValuesOnly ? prefix : encoder(prefix, defaults.encoder, charset, 'key');
+                var commaValue = obj.map(function (item) {
+                    return formatter(encoder(item, defaults.encoder, charset, 'value'));
+                });
+                return [formatter(commaKey) + '=' + commaValue];
+            }
+            return [formatter(prefix) + '=' + formatter(String(obj))];
+        }
+    }
+
     var values = [];
 
     if (typeof obj === 'undefined') {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by ljharb** [comment]
When the format is "RFC3986", this seems correct. However (see https://github.com/ljharb/qs#rfc-3986-and-rfc-1738-space-encoding) when the format is RFC1738, for example, i would not expect it to be encoded.

**Comment by Om4ar** [comment]
will try to do it for RFC3986 only

File: lib/stringify.js [hunk 1]
```diff
@@ -94,6 +92,21 @@ var stringify = function stringify(
         return [formatter(prefix) + '=' + formatter(String(obj))];
     }
 
+    if (generateArrayPrefix === 'comma' && isArray(obj)) {
+        if (formatter !== formats.formatters['RFC3986']) {
+            obj = obj.join(',');
+        } else {
+            if (encoder) {
+                var commaKey = encodeValuesOnly ? prefix : encoder(prefix, defaults.encoder, charset, 'key');
+                var commaValue = obj.map(function (item) {
```
ljharb: we can't add a dependency on `.map` here.

ljharb (changes requested):
I've rebased this and tweaked some things, but an easy solution hasn't yet presented itself.

lib/stringify.js
petersolopov: what is wrong with the `.map`?

ljharb: @petersolopov Pre-ES5 browsers won't have it. You can rebase and use `utils.maybeMap` instead.

ljharb:
@Om4ar are you no longer interested in completing this PR? Please leave it open even so, so that someone else can complete it.

PR #339, opened 2019-10-21
Remove all question marks while ignoreQueryPrefix enabled

webdevium:
Some helper for inattentive developers. Use case taken from real development.

### DIFF
--- a/test/parse.js
+++ b/test/parse.js
@@ -367,6 +367,7 @@ test('parse()', function (t) {
 
     t.test('allows for query string prefix', function (st) {
         st.deepEqual(qs.parse('?foo=bar', { ignoreQueryPrefix: true }), { foo: 'bar' });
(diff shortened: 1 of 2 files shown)


test/parse.js
```diff
@@ -367,6 +367,7 @@ test('parse()', function (t) {
 
     t.test('allows for query string prefix', function (st) {
         st.deepEqual(qs.parse('?foo=bar', { ignoreQueryPrefix: true }), { foo: 'bar' });
+        st.deepEqual(qs.parse('???foo=bar', { ignoreQueryPrefix: true }), { foo: 'bar' });
```
ljharb: This doesn't seem like the right test. At most, it should *throw* if it encounters > 0 question marks, or > 1 when ignoreQueryPrefix (but that'd be a semver-major change).

I'm pretty sure question marks after the first one are a valid part of the query string, that the server can receive.

webdevium: Ok. Thanks for review :)

PR #356, opened 2020-02-25
Exclude unnecessary files from the npm package

pkuczynski:
Listed a few additional files and folders which are currently included in the published npm package, while they should not be.

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,4 +1,9 @@
-bower.json
-component.json
+.github
+test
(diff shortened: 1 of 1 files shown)


pkuczynski:
Is `.travis.yml` (and few others) required too?

ljharb:
As has been discussed on others of your PRs on other packages, that one could be excluded, but what’s the value in doing so for such a tiny file?

PR #358, opened 2020-03-12
[Breaking] add native ESM entry point

d-fischer:
This fixes the following error when importing using the native ESM loader in node:

```
import { stringify } from 'qs';
         ^^^^^^^^^
SyntaxError: The requested module 'qs' does not provide an export named 'stringify'
```

It should also enable tree shaking.


package.json
```diff
@@ -11,6 +11,12 @@
         "url": "https://github.com/sponsors/ljharb"
     },
     "main": "lib/index.js",
+    "exports": {
+        ".": {
```
ljharb: this needs the array fallback to the string form, or it won't work on node 13.0 or 13.1.

package.json
```diff
@@ -11,6 +11,12 @@
         "url": "https://github.com/sponsors/ljharb"
     },
     "main": "lib/index.js",
+    "exports": {
```
ljharb: adding "exports" is a breaking change. this should also include the ability to access package.json.

lib/index.mjs
```diff
@@ -0,0 +1,17 @@
+'use strict';
+
+import stringify from './stringify.js';
+import parse from './parse.js';
+import formats from './formats.js';
+
+export default {
```
ljharb: there's no value or point in the default export if you have the names :-)

ljharb:
Note that you can already get the benefits of "tree shaking" by directly importing `qs/lib/parse` or `qs/lib/stringify`; "tree shaking" only provides benefits when importing more than you need in the first place.

ljharb:
I've gone ahead and updated the PR to be mergeable, but I'm going to hold off until something else forces a v7.

lib/index.mjs
d-fischer: I was thinking that people could do this, like they currently can with the help of bundlers:

```js
import qs from 'qs';

qs.stringify('...');
```

And I think this PR might actually not be a breaking change because of this.

package.json
d-fischer: Can you show me a documentation of this so I can implement it in the PR? I actually haven't heard about this.

d-fischer:
> 
> 
> Note that you can already get the benefits of "tree shaking" by directly importing `qs/lib/parse` or `qs/lib/stringify`; "tree shaking" only provides benefits when importing more than you need in the first place.

I think the usual best practice is that you only import from the entry point, which generally includes everything.

package.json
d-fischer: People actually access other people's package.json and assume it's a part of the public API of packages? Never heard of that

ljharb: Everything reachable is public, yes - and the lack of package.json access broke preact users recently.

ljharb:
That’s the usual practice, yes, but it’s not “best” whatsoever :-)

lib/index.mjs
```diff
@@ -0,0 +1,11 @@
+'use strict';
+
+import stringify from './stringify.js';
+import parse from './parse.js';
+import formats from './formats.js';
+
```
ndelangen, line 6:
```suggestion

export default {
    formats,
    parse,
    stringify
}

```

LeoDog896:
Forgot I even commented in here -- I don't see much benefit to this *other* than the small benefit of tree-shaking by default, and being able to use the native ESM loader (forwards compatibility), so that way you don't need to do:

```
import qs from "qs";
const { parse } = qs
```,

**but** in most cases this shouldn't be an issue. However, I don't think there are many costs to including this in the next breaking change.

**Comment by ljharb** [comment]
tree-shaking is something that can be done just as well in CJS as in ESM, and you can already both require or import `qs/parse` or `qs/stringify`, and do a better job than treeshaking can.

**Comment by ndelangen** [comment]
@ljharb after looking closer, I agree with you this PR actually has no benefits. Or at least not for my usage.

Thank you for all your hard work ❤️ 

What I would be keen to see is version of this library purely for modern environments (chrome100 and node16) which is ESM.
We're using it in Storybook in a few places, and want to start supporting modern-only stuff, and leave the past behind us.

Do you have some roadmap to at some point make a modern-only version, or accept a PR to do it?

@LeoDog896 if you're interested there's a 'tool' to convert any lib to ESM:
https://esm.sh/v86/qs@6.10.5/es2022/qs.js

That's what `qs` looks like completely converted to ESM and bundled (thanks esbuild!)

File: package.json (hunk 2 above)
Comment (drippysip) [reply package.json:14]:
```suggestion
    "exports": [{](url)
```

ljharb:
Given that anybody that can import ESM can also import CJS (in some non-node environments, via a bundler or a transpiler), I don't think there's any benefit in (anyone) shipping native ESM in a package. I'm going to close this.

PR #361, opened 2020-03-25
[Fix] proper comma parsing of URL-encoded comma

### LINKED ISSUE #311: Comma-delimited array parsing fails when individual items contain commas

ljharb:
Followup PR to #336. Fixes #311.

### DIFF
--- a/lib/parse.js
+++ b/lib/parse.js
@@ -37,6 +37,17 @@ var parseArrayValue = function (val, options) {
     return val;
 };
 
+var maybeMap = function maybeMap(val, fn) {
+    if (isArray(val)) {
+        var mapped = [];
+        for (var i = 0; i < val.length; i += 1) {
+            mapped.push(fn(val[i]));
+        }
+        return mapped;
+    }
+    return fn(val);
+};
+
 // This is what browsers will submit when the ✓ character occurs in an
 // application/x-www-form-urlencoded body and the encoding of the page containing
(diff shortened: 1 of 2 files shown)


ljharb:
cc @quentinus95; how does this look?

PR #363, opened 2020-04-29
Add support for stringifying space, pipe, and tab delimited arrays

kylerjensen:
I use this library for stringifying query strings in an API client generated from an OpenAPI spec. As part of the OpenAPI specification, developers may specify that an array may be delimited with a `comma`, `space`, `pipe`, or `tab` character, as well as repeated or exploded. This commit adds support for stringifying arrays with the additional delimiter types.

README, CHANGELOG, test/, and dist/ have all been updated in this PR as well.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+## [Unreleased]
+- [New] Add support stringifying `space`, `pipe`, and `tab` delimited arrays (#363)
+
 ## **6.9.3**
 - [Fix] proper comma parsing of URL-encoded commas (#361)
 - [Fix] parses comma delimited array while having percent-encoded comma treated as normal text (#336)
--- a/README.md
+++ b/README.md
@@ -381,6 +381,12 @@ qs.stringify({ a: ['b', 'c'] }, { arrayFormat: 'repeat' })
 // 'a=b&a=c'
 qs.stringify({ a: ['b', 'c'] }, { arrayFormat: 'comma' })
 // 'a=b,c'
+qs.stringify({ a: ['b', 'c'] }, { arrayFormat: 'space' })
+// 'a=b c'
+qs.stringify({ a: ['b', 'c'] }, { arrayFormat: 'pipe' })
+// 'a=b|c'
+qs.stringify({ a: ['b', 'c'] }, { arrayFormat: 'tab' })
+// 'a=b\tc'
 ```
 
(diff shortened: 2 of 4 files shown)


dist/qs.js
```diff
@@ -324,9 +324,12 @@ var arrayPrefixGenerators = {
     indices: function indices(prefix, key) {
         return prefix + '[' + key + ']';
     },
+    pipe: 'pipe',
```
ljharb, line 327: please revert all changes to this file; it's only updated at release time.

Kjens93: > If we're going to add these options to stringify, we'd need them in `parse` as well, so things can round trip.

@ljharb No problem, I'll make that change. Thank you again for taking the time to review this PR!

Kjens93:
> If we're going to add these options to stringify, we'd need them in `parse` as well, so things can round trip.

@ljharb No problem, I'll make that change. Thank you again for taking the time to review this PR!

Kjens93:
> I'm not convinced that these use cases are generic enough to need first-class support; I wonder if it would be better to come up with a custom option that lets you provide your own delimiter?

@ljharb I absolutely understand and agree. I think a solution that allows for the caller to specify a delimiter is probably the better experience. However, in this PR I wanted to make changes that were minimally invasive and didn't break existing functionality (I felt like the comma functionality should probably use the new mechanism as well if we were going to go down that route). I'm happy to go ahead and make the change you suggested about adding support for the `parse` function, and then if you'd like to fold these into a new major release that supports a custom delimiter at some point I'll leave that up to you. Is that an acceptable solution?

dist/qs.js
Kjens93: Will do 👍

ljharb:
@Kjens93 sorry for the long delay :-)

Adding a new option for a custom delimiter, that existing comma support is refactored to build on top of, should be semver-minor, and I think that's a better approach than adding hardcoded one-offs.

For example, `arrayFormat: 'comma'` could be shorthand for `arrayFormat: 'delimited', delimiter: ','` , and `arrayFormat: 'delimited'` would throw if `delimiter` was not provided. Thoughts?

ljharb:
oof, not sure what happened here.

I guess you deleted your fork? That makes PRs unrecoverable, unfortunately. @kylerjensen if you're still interested in landing this, please file it again on a fork that remains until the PR is merged :-)

PR #365, opened 2020-05-02
Custom stringify arrays and objects

leninlin:
I added the ability to compile the query string using a custom function for arrays and objects.

I need this to work with the Python REST framework (https://github.com/AltSchool/dynamic-rest):
```javascript
qs.stringify({
  include: [ 'name', 'age' ],
  filter: {
    'age.gte': 18
  }
});
// include[]=name&include[]=age&filter{age.gte}=18
```

I added a function with curly brackets to the default list, but custom functions for the future will be useful for such situations.


dist/qs.js
```diff
@@ -329,6 +329,18 @@ var arrayPrefixGenerators = {
     }
 };
 
+var objectPrefixGenerators = {
```
ljharb, line 332: please revert all changes to this file; it's only updated at release time.

ljharb (changes requested):
Any changes to `stringify` would need to be mirrored in `parse`, so things can round trip.

I'm not yet convinced that this complexity is worth adding, but I appreciate you've made it appropriately generic.

dist/qs.js
leninlin: Done

leninlin:
> Any changes to stringify would need to be mirrored in parse, so things can round trip.

I'll think about it now

> I'm not yet convinced that this complexity is worth adding, but I appreciate you've made it appropriately generic.

No problems. I can use fork for myself.

leninlin:
Maybe something like that?

lib/parse.js
```diff
@@ -162,8 +184,18 @@ var parseKeys = function parseQueryStringKeys(givenKey, val, options, valuesPars
         return;
     }
 
-    // Transform dot notation to bracket notation
-    var key = options.allowDots ? givenKey.replace(/\.([^.[]+)/g, '[$1]') : givenKey;
+    // Transforms key to bracket notation
+    var key = givenKey;
+    if (typeof options.arrayFormat === 'function') {
+        key = options.arrayFormat(key);
+    } else if (options.arrayFormat in arrayKeyReplacer) {
```
ljharb, line 191:
```suggestion
    } else if (has.call(arrayKeyReplacer, options.arrayFormat)) {
```

lib/parse.js
```diff
@@ -162,8 +184,18 @@ var parseKeys = function parseQueryStringKeys(givenKey, val, options, valuesPars
...
+        key = arrayKeyReplacer[options.arrayFormat](key);
+    }
+    if (typeof options.objectFormat === 'function') {
+        key = options.objectFormat(key);
+    } else if (options.objectFormat in objectKeyReplacer) {
```
ljharb, line 196:
```suggestion
    } else if (has.call(objectKeyReplacer, options.objectFormat)) {
```

lib/parse.js
```diff
@@ -5,18 +5,40 @@ var utils = require('./utils');
...
+    }
+};
+
+var objectKeyReplacer = {
+    brackets: function brackets(key) {
+        return key;
+    },
+    curly: function curly(key) {
+        return key.replace(/\{([^{}[]+)\}/g, '[$1]');
+    },
```
ljharb, line 20:
```suggestion
```

lib/stringify.js
```diff
@@ -220,16 +233,27 @@ module.exports = function (object, opts) {
...
     } else {
-        arrayFormat = 'indices';
+        generateArrayPrefix = arrayPrefixGenerators.indices;
     }
 
-    var generateArrayPrefix = arrayPrefixGenerators[arrayFormat];
+    var generateObjectPrefix;
+    if (opts && typeof opts.objectFormat === 'function') {
+        generateObjectPrefix = opts.objectFormat;
+    } else if (opts && opts.objectFormat in objectPrefixGenerators) {
```
ljharb, line 250:
```suggestion
    } else if (opts && has.call(objectPrefixGenerators, opts.objectFormat)) {
```

lib/stringify.js
```diff
@@ -220,16 +233,27 @@ module.exports = function (object, opts) {
...
+        generateObjectPrefix = objectPrefixGenerators[opts.objectFormat];
+    } else if (opts && opts.allowDots) {
+        generateObjectPrefix = objectPrefixGenerators.dots;
+    } else {
+        generateObjectPrefix = objectPrefixGenerators.brackets;
+    }
```
ljharb, line 256: this seems like it should be handled by the option normalization and the default?

test/parse.js
```diff
@@ -40,36 +40,81 @@ test('parse()', function (t) {
...
         st.deepEqual(qs.parse('a=b&a=c', { arrayFormat: 'indices' }), { a: ['b', 'c'] });
         st.end();
     });
 
-    t.test('arrayFormat: comma allows only comma-separated arrays', function (st) {
+    t.test('arrayFormat: comma allows comma-separated arrays', function (st) {
         st.deepEqual(qs.parse('a[]=b&a[]=c', { arrayFormat: 'comma' }), { a: ['b', 'c'] });
         st.deepEqual(qs.parse('a[0]=b&a[1]=c', { arrayFormat: 'comma' }), { a: ['b', 'c'] });
-        st.deepEqual(qs.parse('a=b,c', { arrayFormat: 'comma' }), { a: 'b,c' });
+        st.deepEqual(qs.parse('a=b,c', { arrayFormat: 'comma' }), { a: ['b', 'c'] });
```
ljharb, line 54: this shouldn't change, or else it's a breaking change

ljharb (changes requested):
I'm confused why there's both an array replacer and an object replacer, since for parse, they both do the same thing.

ljharb:
@leninlin are you still interested in completing this PR?

PR #389, opened 2021-01-01
Only publish required files

willfarrell:
See https://docs.npmjs.com/cli/v6/configuring-npm/package-json#files for details. Every little bit to improve cold start times.

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,6 +11,10 @@
         "url": "https://github.com/sponsors/ljharb"
     },
     "main": "lib/index.js",
+    "files":[
+        "dist",
+        "lib"
+    ],
     "contributors": [
         {
             "name": "Jordan Harband",


package.json
```diff
@@ -11,6 +11,10 @@
         "url": "https://github.com/sponsors/ljharb"
     },
     "main": "lib/index.js",
+    "files":[
+        "dist",
+        "lib"
+    ],
```
ljharb, line 17: The `files` field is dangerous; it means needed files could be accidentally omitted. The only danger of `.npmignore` is that an unneeded file is unnecessarily included.

ljharb (changes requested):
All files are required - `npm explore foo && npm install && npm test` should always work.###

ljharb:
Duplicate of #356. Duplicate of #250.

The only files that currently show up in `npm pack --dry-run` that *could* be excluded are `.github/workflows/rebase.yml` and `.github/FUNDING.yml ` - every other file is 100% necessary in the published package.

ljharb:
Separately, "cold start time" is a runtime concern, and excluding files from the published tarball has no bearing on cold start times - only (imperceptibly) on install times.

package.json
willfarrell: Sorry, missed the npm ignore file.

ljharb: np, unfortunately npm's own docs recommend "files", but it's the more dangerous approach.

willfarrell:
Thanks for the quick reply and detailed explanation. Everyone should be bundling their code in which case it doesn't have an impact on cold start time as you stated above. However, this isn't always the case and people package up everything and ship it. Will an extra 1-3kb make a difference, of course not, but when 100s of packages have extra file, they can add up to MB which can impact cold starts in my experience.

I'm mostly trying to micro optimize for those less experienced developers that end up using [middy.js v2](https://github.com/middyjs/middy/issues/585). Thanks for taking the time to review.

ljharb:
I'm not talking about bundling - files on disk that are never required hav e *zero* impact on start time, because they're never touched.

It is, quite simply, impossible for a cold start to be impacted by unused files, unless you're including `npm install` time, or "upload and extract the deploy artifact" in start time, neither of which is typically included in that metric.

PR #393, opened 2021-01-27
[New] `stringify`: throw on cycles, instead of an infinite loop

### LINKED ISSUE #367: Throw error on circular references
Currently the library does not throw error when parsing circual references.

Example:

```javascript
const circular = {
   a : "value"
}
circular.a = circular
JSON.stringify(circular) //throws error
```

```javascript
const circular = {
   a : "value"
}
circular.a = circular
qs.stringify(circular) //freezes computer 😣
```

Poyoman39:
Fix issue #367 with "maxDepth" option.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -10,7 +10,7 @@
         "id-length": [2, { "min": 1, "max": 25, "properties": "never" }],
         "indent": [2, 4],
         "max-lines-per-function": [2, { "max": 150 }],
-        "max-params": [2, 14],
+        "max-params": [2, 15],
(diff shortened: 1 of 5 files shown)


ljharb (changes requested):
Thanks!

We'd definitely need tests for this; but i'm not sure a max depth is a proper fix for the solution - typically it's done by recognizing "seen objects" in some kind of WeakMap/Map/array solution.

Poyoman39:
Ok i will provide a Map solution ;)

PR #395, opened 2021-01-28
[New] `stringify`: throw on cycles, instead of an infinite loop

### LINKED ISSUE #367: Throw error on circular references

Poyoman39:


ljharb (approved):
updated all three PRs to use `side-channel`

PR #408, opened 2021-04-09
[Docs] add note and links for coercing primitive values

DesignByOnyx:
This should hopefully lay to rest #91

### DIFF
--- a/README.md
+++ b/README.md
@@ -287,6 +287,17 @@ assert.deepEqual(arraysOfObjects, { a: ['b', 'c'] })
 ```
 (_this cannot convert nested objects, such as `a={b:1},{c:d}`_)
 
+### Parsing primitive/scalar values (numbers, booleans, null, etc)
+
+By default, all values are parsed as strings. This behavior will not change and is explained in [issue #91](https://github.com/ljharb/qs/issues/91).
+
+```javascript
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -287,6 +287,20 @@ assert.deepEqual(arraysOfObjects, { a: ['b', 'c'] })
...
+
+```javascript
+var primitiveValues = qs.parse('a=15&b=true&c=null');
+assert.deepEqual(primitiveValues, { a: '15', b: 'true', c: 'null' });
+```
+
+If you wish to auto-convert values which look like numbers, booleans, and other values into their primitive values, you can do one of the following:
+
+-   [Install Express JS middleware to auto-convert request query parameters](https://github.com/xpepermint/query-types)
+-   [Copy the above MIT licensed coercion code](https://github.com/xpepermint/query-types/blob/master/index.js) directly into your project and modify it as you see fit. This code has been used on some very large projects and gives you the greatest amount of control over your data.
```
ljharb, line 302: let's not advise people to copypasta code; if they want to use that code, they should depend on the package.

DesignByOnyx: Done - thanks for the feedback.

ljharb (approved):

PR #424, opened 2021-11-05
Fixes the issue of stringify encoding comma(used to join array) even when `encodeValuesOnly = true`

adnan-creator:
Following changes were made:
* According to the documentation when `encodeValuesOnly = true`, then all characters except the values must not be encoded. This includes brackets and joining characters such as comma. It works perfectly for all other characters except when comma is used to join elements of an array. Then even the comma is encoded. 
* One of the tests had a bug where we assumed the result of `qs.stringify({ a: ['b', 'c', 'd'] }, { arrayFormat: 'comma' })` to be equal to `'a=b,c,d'`. Which is not the expected behavior as for the comma to not be encoded we must either set `encode: false` or `encodeValuesOnly: true`. So the right result would be `'a=b%2Cc%2Cd'` as the default functionality of the function is to encode the complete string.

### DIFF
--- a/lib/stringify.js
+++ b/lib/stringify.js
@@ -19,6 +19,7 @@ var arrayPrefixGenerators = {
 };
 
 var isArray = Array.isArray;
+var split = String.prototype.split;
 var push = Array.prototype.push;
 var pushToArray = function (arr, valueOrArray) {
     push.apply(arr, isArray(valueOrArray) ? valueOrArray : [valueOrArray]);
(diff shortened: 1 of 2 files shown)


lib/stringify.js
```diff
@@ -102,6 +102,14 @@ var stringify = function stringify(
     if (isNonNullishPrimitive(obj) || utils.isBuffer(obj)) {
         if (encoder) {
             var keyValue = encodeValuesOnly ? prefix : encoder(prefix, defaults.encoder, charset, 'key', format);
+            if(generateArrayPrefix === 'comma' && encodeValuesOnly){
```
ljharb, line 105: nit:
```suggestion
            if (generateArrayPrefix === 'comma' && encodeValuesOnly) {
```

lib/stringify.js
```diff
@@ -102,6 +102,14 @@ var stringify = function stringify(
...
+                const valuesArray = String(obj).split(',');
```
ljharb, line 106: `String(obj)` for a non-array object will turn into `[object Object]`, which i don't think is what we want.

ljharb (changes requested):
I rebased this on top of my failing tests branch; take a look at the failures. The `???` will need to be updated to an actual expected value.

ljharb (approved):
Overall this looks great!

Are there any additional test cases we should be adding to make sure this is doing both what the issue participants expect, and what we expect?

adnan-creator:
Thanks! for now these tests seem fine.

ljharb (approved):

PR #426, opened 2021-12-04
[Fix] `stringify`: actually fix cyclic references

### LINKED ISSUE #403: qs.stringify Uncaught RangeError: Cyclic object value
var qs = require('qs');
var a = ['a'];
qs.stringify({x:a,y:a});

-> Uncaught RangeError: Cyclic object value

liaokunhua:
With all the testing, I think this is the ultimate solution for circular references.

Fixes #403.

### DIFF
--- a/lib/stringify.js
+++ b/lib/stringify.js
@@ -56,6 +56,8 @@ var isNonNullishPrimitive = function isNonNullishPrimitive(v) {
         || typeof v === 'bigint';
 };
 
+var sentinel = {};
+
 var stringify = function stringify(
     object,
     prefix,
@@ -75,8 +77,23 @@ var stringify = function stringify(
 ) {
     var obj = object;
 
-    if (sideChannel.has(object)) {
-        throw new RangeError('Cyclic object value');
+    var tmpSc = sideChannel;
+    var step = 0;
+    var findFlag = false;
+    while ((tmpSc = tmpSc.get(sentinel)) !== undefined && !findFlag) {
+        // Where object last appeared in the ref tree
+        var pos = tmpSc.get(object);
+        step += 1;
+        if (typeof pos !== 'undefined') {
+            if (pos === step) {
+                throw new RangeError('Cyclic object value');
+            } else {
+                findFlag = true; // Break while
+            }
+        }
+        if (typeof tmpSc.get(sentinel) === 'undefined') {
+            step = 0;
+        }
     }
 
     if (typeof filter === 'function') {
@@ -145,8 +162,9 @@ var stringify = function stringify(
             ? typeof generateArrayPrefix === 'function' ? generateArrayPrefix(prefix, key) : prefix
             : prefix + (allowDots ? '.' + key : '[' + key + ']');
 
-        sideChannel.set(object, true);
+        sideChannel.set(object, step);
         var valueSideChannel = getSideChannel();
+        valueSideChannel.set(sentinel, sideChannel);
         pushToArray(values, stringify(
             value,
             keyPrefix,
--- a/test/stringify.js
+++ b/test/stringify.js
@@ -454,7 +454,7 @@ test('stringify()', function (t) {
 
         st['throws'](
             function () { qs.stringify({ 'foo[bar]': 'baz', 'foo[baz]': a }); },
-            RangeError,
+            /RangeError: Cyclic object value/,
             'cyclic values throw'
         );
 
@@ -464,8 +464,14 @@ test('stringify()', function (t) {
         circular.a = circular;
         st['throws'](
             function () { qs.stringify(circular); },
-            RangeError,
+            /RangeError: Cyclic object value/,
             'cyclic values throw'
+        );
+
+        var arr = ['a'];
+        st.doesNotThrow(
+            function () { qs.stringify({ x: arr, y: arr }); },
+            'non-cyclic values do not throw'
         );
 
         st.end();


lib/stringify.js
```diff
@@ -56,6 +56,8 @@ var isNonNullishPrimitive = function isNonNullishPrimitive(v) {
         || typeof v === 'bigint';
 };
 
+var scSym = Symbol('SideChannel');
```
ljharb, line 59: We support many engines that do not have symbols.

lib/stringify.js
```diff
@@ -75,8 +77,23 @@ var stringify = function stringify(
...
-    if (sideChannel.has(object)) {
-        throw new RangeError('Cyclic object value');
+    var tmpSc = sideChannel;
+    var step = 0;
+    var findFlag = false;
+    while ((tmpSc = tmpSc.get(scSym)) !== undefined && !findFlag) {
+        // Where object last appeared in the ref tree
+        var pos = tmpSc.get(object);
+        step += 1;
+        if (pos != undefined) {
```
ljharb, line 87: This should use !==, or != null; != should never be used to compare to anything besides null.

ljharb (changes requested):
Can you provide a regression test that this fixes?

liaokunhua:
> Can you provide a regression test that this fixes?

Sorry bro, this is a bit out of my league, I haven't thought of a special object that would prove i didn't fix this bug. 

So, let's talk about how I found this bug.
I started learning front-end and github a few days ago, starting with a simple project.
I noticed that the stringify.js file in the test report had a line of code that was not covered.

=> `throw new RangeError('Cyclic object value');`

I did some tests, the result is that non-circular duplicated references can work, but circular dependencies result in "RangeError: Maximum call stack size exceeded", not "RangeError: Cyclic object value".

I think the exception thrown in the use case test of the circular dependency happens to be RangeError, but the exception message is not checked.

lib/stringify.js
liaokunhua: emmm....it was my poor consideration, I meant to define a unique symbol to identify the side-channel, and Symbol is good idea.

lib/stringify.js
liaokunhua: i get it, thanks for the guidance.

liaokunhua: > 

i get it, thanks for the guidance.

ljharb:
Surely you provided some input that caused that error?

lib/stringify.js
```diff
@@ -56,7 +56,7 @@ var isNonNullishPrimitive = function isNonNullishPrimitive(v) {
         || typeof v === 'bigint';
 };
 
-var scSym = Symbol('SideChannel');
+var scSym = Object.create(null);
```
ljharb, line 59: Unfortunately Object.create also can’t be relied upon.
```suggestion
var sentinel = {};
```

ljharb (changes requested):

liaokunhua:
> Surely you provided some input that caused that error?

Right, an example of a circular reference is constructed, which is inside the use case test.
It is easy to reproduce.

Based on the original code for testing.
Step1: 
modify line 79 in the stringify.js file, `throw new RangeError('Cyclic object value');` to `throw new Error('Cyclic object value');`.

Step2:
Create a new test and select an example of a circular reference in the stringify.js file inside the test folder and make the changes.
```
t.test("does not crash when parsing circular references", function(st) {
        var circular = { a: 'value' };
        circular.a = circular;
        try {
            qs.stringify(circular);
        } catch(e) {
            console.dir(e)
        }
        st['throws'](
            function () { qs.stringify(circular); },
            RangeError,
            'Cyclic values throw'
        );
        st.end();
})
```

Step3:
Run test and get output:
```
TAP version 13
# stringify()
# does not crash when parsing circular references
RangeError: Maximum call stack size exceeded
// ....omitted stack info
ok 1 Cyclic values throw

1..1
# tests 1
# pass  1

# ok
```

I was dumbfounded!! not the expected "Error:Cyclic object value".

Environment information:
system: ubuntu 20.04
node: v14.16.1
npm: 6.14.12

ljharb:
ah, i see - so changing that test to check the error message - so we know it's not a stack overflow - would reproduce the error that this PR fixes?

if so, could you make that change?

liaokunhua:
> ah, i see - so changing that test to check the error message - so we know it's not a stack overflow - would reproduce the error that this PR fixes?
> 
> if so, could you make that change?

emmm...kinda wish it was me, but i don't think i have it in me yet.😅

lib/stringify.js
liaokunhua: > 

Submitted new code. for this question, i didn't understand why it was unreliable. two objects created by `Object.create`, return value is false when judged by ===.

liaokunhua: I see, polyfill is provided in MDN and it doesn't support null as the first parameter, is that the reason?

ljharb: It's impossible to polyfill null objects.

We support down to IE 6, which doesn't have Object.create or `__proto__`.

ljharb:
I rebased this PR, and updated the test cases in the default branch that were asserting a RangeError, but not the right one, and added another test case that confirms a fix for #403.

lib/stringify.js
```diff
@@ -75,8 +77,23 @@ var stringify = function stringify(
 ) {
     var obj = object;
 
-    if (sideChannel.has(object)) {
-        throw new RangeError('Cyclic object value');
+    var tmpSc = sideChannel;
+    var step = 0;
+    var findFlag = false;
+    while ((tmpSc = tmpSc.get(sentinel)) !== undefined && !findFlag) {
```
ljharb, line 83: is there any way to do this without a loop, or at least without doing an assignment inside the conditional expression?

ljharb (approved):
actually i'll merge this as-is, to fix the bug; but i'd love it if there was a followup to clean up the loop.

lib/stringify.js
liaokunhua: no other graceful way come to mind at the moment, and if I think of any later, I will submit another PR.

PR #427, opened 2021-12-27
[Robustness] `stringify`: avoid relying on a global `undefined`

Connormiha:
_I just removed needless(for these cases) `typeof` for checking for undefined_

update: Added checking for typeof in parser.

### DIFF
--- a/lib/stringify.js
+++ b/lib/stringify.js
@@ -80,7 +80,7 @@ var stringify = function stringify(
     var tmpSc = sideChannel;
     var step = 0;
     var findFlag = false;
-    while ((tmpSc = tmpSc.get(sentinel)) !== undefined && !findFlag) {
+    while ((tmpSc = tmpSc.get(sentinel)) !== void undefined && !findFlag) {
         // Where object last appeared in the ref tree
         var pos = tmpSc.get(object);
         step += 1;
@@ -142,7 +142,7 @@ var stringify = function stringify(
     var objKeys;
     if (generateArrayPrefix === 'comma' && isArray(obj)) {
         // we need to join elements in
-        objKeys = [{ value: obj.length > 0 ? obj.join(',') || null : undefined }];
+        objKeys = [{ value: obj.length > 0 ? obj.join(',') || null : void undefined }];
     } else if (isArray(filter)) {
         objKeys = filter;
     } else {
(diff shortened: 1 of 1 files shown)


ljharb (changes requested):
These are not needless at all. `undefined` can be redefined in every scope except the global scope in ES5, and in ES3 - which this package supports - it can be modified in the global scope otherwise.

Connormiha:
@ljharb 
> These are not needless at all. `undefined` can be redefined in every scope except the global scope in ES5, and in ES3 - which this package supports - it can be modified in the global scope otherwise.

Yes, but there are many places with same check. For example https://github.com/ljharb/qs/blob/master/lib/stringify.js#L155 or https://github.com/ljharb/qs/blob/master/lib/stringify.js#L195

ljharb:
In that case, those should be changed to use typeof instead :-) want to update the PR to do that?

Connormiha:
@ljharb `typeof` or `void 0`? Which one is preferable?

Connormiha:
@ljharb updated.

ljharb:
Definitely the `typeof`, because it has minimal runtime weight.

lib/stringify.js
```diff
@@ -80,7 +80,7 @@ var stringify = function stringify(
     var tmpSc = sideChannel;
     var step = 0;
     var findFlag = false;
-    while ((tmpSc = tmpSc.get(sentinel)) !== undefined && !findFlag) {
+    while (typeof (tmpSc = tmpSc.get(sentinel)) !== 'undefined' && !findFlag) {
```
ljharb, line 83: altho this one is actually a bit trickier because of the assignment in the loop, so let's do this:
```suggestion
    while ((tmpSc = tmpSc.get(sentinel)) !== void undefined && !findFlag) {
```

lib/stringify.js
```diff
@@ -142,7 +142,7 @@ var stringify = function stringify(
     var objKeys;
     if (generateArrayPrefix === 'comma' && isArray(obj)) {
         // we need to join elements in
-        objKeys = [{ value: obj.length > 0 ? obj.join(',') || null : undefined }];
+        objKeys = [{ value: obj.length > 0 ? obj.join(',') || null : (void 0) }];
```
ljharb, line 145:
```suggestion
        objKeys = [{ value: obj.length > 0 ? obj.join(',') || null : void undefined }];
```
since it's a bit clearer

ljharb (commented):
Thanks! Almost there.

ljharb (approved):
Thanks!

PR #400, opened 2021-03-15
[meta] Add test and configuration files to .npmignore

ixth:
I think most users don't need neither tests nor config files when they download a package. I'd also remove CHANGELOG.md, but npm won't let me do that.

Package size reduced by ~40% both packed and unpacked.

```diff
@@ -9,34 +9,26 @@ Detected non-install and non-publish: exiting.
 npm notice
 npm notice 📦  qs@6.9.6
 npm notice === Tarball Contents ===
-npm notice 520B   .editorconfig
-npm notice 16B    .eslintignore
-npm notice 1.0kB  .eslintrc
-npm notice 216B   .nycrc
 npm notice 476B   lib/formats.js
 npm notice 211B   lib/index.js
 npm notice 9.3kB  lib/parse.js
-npm notice 33.3kB test/parse.js
 npm notice 26.1kB dist/qs.js
 npm notice 8.4kB  lib/stringify.js
-npm notice 30.8kB test/stringify.js
 npm notice 6.8kB  lib/utils.js
-npm notice 5.1kB  test/utils.js
 npm notice 2.0kB  package.json
 npm notice 20.0kB CHANGELOG.md
 npm notice 1.6kB  LICENSE.md
 npm notice 19.8kB README.md
-npm notice 548B   .github/FUNDING.yml
 npm notice === Tarball Details ===
 npm notice name:          qs
 npm notice version:       6.9.6
 npm notice filename:      qs-6.9.6.tgz
-npm notice package size:  37.0 kB
-npm notice unpacked size: 168.4 kB
-npm notice shasum:        af29cab02a478195f2e423b84c167d9cd927bd21
-npm notice integrity:     sha512-qF2xgWi49vJRz[...]czs7/dQzToK6g==
-npm notice total files:   20
+npm notice package size:  20.6 kB
+npm notice unpacked size: 96.9 kB
+npm notice shasum:        ef5052808def8c8c09588a370dc814d40a7e8d80
+npm notice integrity:     sha512-LB4f8bQhvX4NT[...]QTl7Y0WesYX6A==
+npm notice total files:   12
 npm notice
 qs-6.9.6.tgz
```

PR #429, opened 2022-01-12
Update .npmignore

udivankin:
exclude .editorconfig .eslintignore .eslintrc from bundling to npm package

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -15,5 +15,8 @@ coverage/
 bower.json
 component.json
 .npmignore
+.editorconfig
+.eslintignore
+.eslintrc
 
 .github/workflows


ljharb (changes requested):
no, these are intentionally included so that `npm explore qs && npm install && npm test` always works.

ljharb:
Duplicate of #400. Duplicate of #389. Duplicate of #356. Duplicate of #250.

> The only files that currently show up in `npm pack --dry-run` that *could* be excluded are `.github/FUNDING.yml ` - every other file is 100% necessary in the published package.

udivankin:
Well, I've read #250 and it still feels a bit weird for me.
What about putting a short notice in .gitignore explaining why these were intentionally opted out

ljharb:
Given that there's an explicit `.npmignore` file, it seems like a reader should expect that there is a reason without any additional comment, instead of assuming that it's a mistake.

The presence of the `.npmignore` file hasn't stopped people from filing PRs with the "files" field, and I suspect a comment in either `.gitignore` or `.npmignore` would also not achieve the desired effect :-/

PR #430, opened 2022-01-20
[utils] Simplify check isRegExp

Connormiha:
I refactored `isRegExp` to use short method for checking.

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -212,7 +212,7 @@ var compact = function compact(value) {
 };
 
 var isRegExp = function isRegExp(obj) {
-    return Object.prototype.toString.call(obj) === '[object RegExp]';
+    return obj instanceof RegExp;
 };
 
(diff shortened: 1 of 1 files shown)


lib/utils.js
```diff
@@ -212,7 +212,7 @@ var compact = function compact(value) {
 };
 
 var isRegExp = function isRegExp(obj) {
-    return Object.prototype.toString.call(obj) === '[object RegExp]';
+    return obj instanceof RegExp;
```
ljharb, line 215: unfortunately this is not short, it's broken. `instanceof` doesn't work across realms, and can be easily forged.

ljharb (changes requested):

ljharb:
Please discuss changes like this in issues first, rather than sending a PR that might not be needed :-) all the "is" checks in this library are precisely what is needed.

PR #433, opened 2022-03-05
fix: allow parsing of empty keys in objects

### LINKED ISSUE #432: qs.parse omitting empty string keys
Hi, it will be nice to have this part configurable:

https://github.com/ljharb/qs/blob/542a5c7ff88d7229efa2e22c7c8a7d69375f5e72/lib/parse.js#L150-L152

current output:
```ts
qs.parse("=1&=2") === {}
qs.parse(" =1& =2") === { " ": ["1","2"] }
```

expected output :
```ts
qs.parse("=1&=2", { allowEmptyKeys: true }) === { "": ["1","2"] }
```

Coobaha:
Resolves #432 

- [x] to add an option to both parse and stringify
- [ ] to have tests, for both parse and stringify, for at least:
  - [x] repeat keys with all array formats
  - [x] bracketed keys with all array formats
  - [x] a lone = (and =& and &= and &=&)
  - [ ] ❌ deep array/object case  

**If we are happy with this implementation and want to proceed:**

 - [x]  to document in the readme
 - [x]  split tests before merge to parse/stringify files
 - [ ] update https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/qs (also `allowSparse` is missing there and maybe other types)
 
 
 Other:
 - Stringify works without an option, so it is potentially a patch but altering parse behaviour only via option seems a safe bet to not accidentally break ppl code, and can be considered as a default behaviour in the next major version?

### DIFF
--- a/lib/parse.js
+++ b/lib/parse.js
@@ -7,6 +7,7 @@ var isArray = Array.isArray;
 
 var defaults = {
     allowDots: false,
+    allowEmptyKeys: false,
     allowPrototypes: false,
     allowSparse: false,
     arrayLimit: 20,
@@ -84,6 +85,9 @@ var parseValues = function parseQueryStringValues(str, options) {
         var key, val;
         if (pos === -1) {
             key = options.decoder(part, defaults.decoder, charset, 'key');
+            if (key === '') {
+                continue;
+            }
             val = options.strictNullHandling ? null : '';
         } else {
             key = options.decoder(part.slice(0, pos), defaults.decoder, charset, 'key');
@@ -149,7 +153,7 @@ var parseObject = function (chain, val, options, valuesParsed) {
 };
 
 var parseKeys = function parseQueryStringKeys(givenKey, val, options, valuesParsed) {
-    if (!givenKey) {
+    if (!givenKey && (!options.allowEmptyKeys || givenKey !== '')) {
         return;
     }
 
@@ -169,7 +173,7 @@ var parseKeys = function parseQueryStringKeys(givenKey, val, options, valuesPars
     // Stash the parent if it exists
 
     var keys = [];
-    if (parent) {
+    if (parent || (options.allowEmptyKeys && parent === '')) {
         // If we aren't using plain objects, optionally prefix keys that would overwrite object prototype properties
         if (!options.plainObjects && has.call(Object.prototype, parent)) {
             if (!options.allowPrototypes) {
@@ -218,6 +222,7 @@ var normalizeParseOptions = function normalizeParseOptions(opts) {
 
     return {
         allowDots: typeof opts.allowDots === 'undefined' ? defaults.allowDots : !!opts.allowDots,
+        allowEmptyKeys: typeof opts.allowEmptyKeys === 'undefined' ? defaults.allowEmptyKeys : !!opts.allowEmptyKeys,
         allowPrototypes: typeof opts.allowPrototypes === 'boolean' ? opts.allowPrototypes : defaults.allowPrototypes,
         allowSparse: typeof opts.allowSparse === 'boolean' ? opts.allowSparse : defaults.allowSparse,
         arrayLimit: typeof opts.arrayLimit === 'number' ? opts.arrayLimit : defaults.arrayLimit,
--- a/test/empty-keys-cases.js
+++ b/test/empty-keys-cases.js
@@ -2,36 +2,36 @@
 
 module.exports = {
     emptyTestCases: [
-        { input: '&', withEmptyKeys: {}, stringifyOutput: '', noEmptyKeys: {} },
-        { input: '&&', withEmptyKeys: {}, stringifyOutput: '', noEmptyKeys: {} },
-        { input: '&=', withEmptyKeys: { '': '' }, stringifyOutput: '=', noEmptyKeys: {} },
-        { input: '&=&', withEmptyKeys: { '': '' }, stringifyOutput: '=', noEmptyKeys: {} },
-        { input: '&=&=', withEmptyKeys: { '': ['', ''] }, stringifyOutput: '[0]=&[1]=', noEmptyKeys: {} },
-        { input: '&=&=&', withEmptyKeys: { '': ['', ''] }, stringifyOutput: '[0]=&[1]=', noEmptyKeys: {} },
+        { input: '&', withEmptyKeys: {}, stringifyOutput: '', noEmptyKeys: {}, stringifyOutputNoEmpty: '' },
+        { input: '&&', withEmptyKeys: {}, stringifyOutput: '', noEmptyKeys: {}, stringifyOutputNoEmpty: '' },
+        { input: '&=', withEmptyKeys: { '': '' }, stringifyOutput: '=', noEmptyKeys: {}, stringifyOutputNoEmpty: '' },
+        { input: '&=&', withEmptyKeys: { '': '' }, stringifyOutput: '=', noEmptyKeys: {}, stringifyOutputNoEmpty: '' },
+        { input: '&=&=', withEmptyKeys: { '': ['', ''] }, stringifyOutput: '[0]=&[1]=', noEmptyKeys: {}, stringifyOutputNoEmpty: '' },
+        { input: '&=&=&', withEmptyKeys: { '': ['', ''] }, stringifyOutput: '[0]=&[1]=', noEmptyKeys: {}, stringifyOutputNoEmpty: '' },
 
-        { input: '=', withEmptyKeys: { '': '' }, noEmptyKeys: {}, stringifyOutput: '=' },
-        { input: '=&', withEmptyKeys: { '': '' }, stringifyOutput: '=', noEmptyKeys: {} },
-        { input: '=&&&', withEmptyKeys: { '': '' }, stringifyOutput: '=', noEmptyKeys: {} },
-        { input: '=&=&=&', withEmptyKeys: { '': ['', '', ''] }, stringifyOutput: '[0]=&[1]=&[2]=', noEmptyKeys: {} },
-        { input: '=&a[]=b&a[1]=c', withEmptyKeys: { '': '', a: ['b', 'c'] }, stringifyOutput: '=&a[0]=b&a[1]=c', noEmptyKeys: { a: ['b', 'c'] } },
-        { input: '=a', withEmptyKeys: { '': 'a' }, noEmptyKeys: {}, stringifyOutput: '=a' },
-        { input: '=a', withEmptyKeys: { '': 'a' }, noEmptyKeys: {}, stringifyOutput: '=a' },
-        { input: 'a==a', withEmptyKeys: { a: '=a' }, noEmptyKeys: { a: '=a' }, stringifyOutput: 'a==a' },
+        { input: '=', withEmptyKeys: { '': '' }, noEmptyKeys: {}, stringifyOutput: '=', stringifyOutputNoEmpty: '' },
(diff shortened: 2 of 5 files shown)


test/parse.js
```diff
@@ -843,3 +843,52 @@ test('parse()', function (t) {
...
+        st.deepEqual(qs.stringify({ '': ['a', 'b'] }), '%5B0%5D=a&%5B1%5D=b');
+        st.deepEqual(qs.stringify({ '': ['a', 'b'], a: [1, 2] }, { encode: false }), '[0]=a&[1]=b&a[0]=1&a[1]=2');
+
+        st.end();
+    });
+    t.test('edge case with object/arrays', function (st) {
+
+        st.deepEqual(qs.stringify({ '': { '': [2, 3] } }, { encode: false }), '[][0]=2&[][1]=3');
+
+        st.deepEqual(qs.parse('[][0]=2&[][1]=3', { allowEmptyKeys: true }), { '': { '': [2, 3] } });
```
Coobaha, line 890: This fails case fails, but this is ambiguous string which is treated as `{ : [ '2', '3' ] }` and stringify <-> parse is not producing the same results

lib/parse.js
```diff
@@ -83,6 +84,9 @@ var parseValues = function parseQueryStringValues(str, options) {
         var key, val;
         if (pos === -1) {
             key = options.decoder(part, defaults.decoder, charset, 'key');
+            if (key === '') {
+                continue;
```
Coobaha, line 88: handles lonely `&`

test/parse.js
```diff
@@ -843,3 +843,53 @@ test('parse()', function (t) {
...
+        st.deepEqual(qs.stringify({ '': ['a', 'b'], a: [1, 2] }, { encode: false }), '[0]=a&[1]=b&a[0]=1&a[1]=2');
+
+        st.end();
+    });
+    t.test('edge case with object/arrays', function (st) {
+
+        st.deepEqual(qs.stringify({ '': { '': [2, 3] } }, { encode: false }), '[][0]=2&[][1]=3');
+        st.deepEqual(qs.stringify({ '': { '': [2, 3], a: 2 } }, { encode: false }), '[][0]=2&[][1]=3&[a]=2');
+        st.deepEqual(qs.parse('[][0]=2&[][1]=3&[a]=2', { allowEmptyKeys: true }), { '': { '': ['2', '3'] } });
+        st.deepEqual(qs.parse('[][0]=2&[][1]=3&[a]=2', { allowEmptyKeys: true }), { '': { '': ['2', '3'], a: '2' } });
```
Coobaha, line 891:
```tape
  ---
    operator: deepEqual
    expected: |-
      { : { : [ '2', '3' ] } }
    actual: |-
      { : { 0: '2', 1: '3', a: '2' } }
```

```
    operator: deepEqual
    expected: |-
      { : { : [ '2', '3' ], a: '2' } }
    actual: |-
      { : { 0: '2', 1: '3', a: '2' } }
```

test/parse.js
```diff
@@ -843,3 +843,53 @@ test('parse()', function (t) {
...
+        st.deepEqual(qs.stringify({ '': ['a', 'b'], a: [1, 2] }, { encode: false }), '[0]=a&[1]=b&a[0]=1&a[1]=2');
+
+        st.end();
+    });
+    t.test('edge case with object/arrays', function (st) {
+
+        st.deepEqual(qs.stringify({ '': { '': [2, 3] } }, { encode: false }), '[][0]=2&[][1]=3');
+        st.deepEqual(qs.stringify({ '': { '': [2, 3], a: 2 } }, { encode: false }), '[][0]=2&[][1]=3&[a]=2');
+        st.deepEqual(qs.parse('[][0]=2&[][1]=3', { allowEmptyKeys: true }), { '': { '': ['2', '3'] } });
+        st.deepEqual(qs.parse('[][0]=2&[][1]=3&[a]=2', { allowEmptyKeys: true }), { '': { '': ['2', '3'], a: '2' } });
```
Coobaha, line 891: This fails case fails, but this is ambiguous string which is treated as { : [ '2', '3' ] } and stringify <-> parse is not producing the same results

```
  ---
    operator: deepEqual
    expected: |-
      { : { : [ '2', '3' ] } }
    actual: |-
     { : [ '2', '3' ] }

```

```
    operator: deepEqual
    expected: |-
      { : { : [ '2', '3' ], a: '2' } }
    actual: |-
      { : { 0: '2', 1: '3', a: '2' } }
```

ljharb:
There's unlikely to be a major version any time soon; breaking changes just aren't worth it. Configuring it via option is the best bet for now.

lib/parse.js
ljharb: what if the key already exists? this would bypass the `utils.combine` logic below

lib/parse.js
```diff
@@ -148,7 +152,7 @@ var parseObject = function (chain, val, options, valuesParsed) {
 };
 
 var parseKeys = function parseQueryStringKeys(givenKey, val, options, valuesParsed) {
-    if (!givenKey) {
+    if (options.allowEmptyKeys ? givenKey !== '' && !givenKey : !givenKey) {
```
ljharb, line 155:
```suggestion
    if (!givenKey && (!options.allowEmptyKeys || givenKey !== '')) {
```
to avoid the inline ternary

test/parse.js
```diff
@@ -843,3 +843,53 @@ test('parse()', function (t) {
 
     t.end();
 });
+
+test('parse/stringify empty key WIP', function (t) {
+    t.test('parses an object with empty string key', function (st) {
+        st.deepEqual(qs.parse('=a', { allowEmptyKeys: true }), { '': 'a' });
```
ljharb, line 849: for each of these test inputs, let's test both allowEmptyKeys true, false, and absent

test/parse.js
ljharb: it's ok to mark this case as a TODO for now; there's an existing issue about a nested array/object combo

ljharb (commented):
At the very least, let's add stringify tests using all of the outputs in the parse tests as input.

test/parse.js
Coobaha: Done, please let me know what will be the correct place for this tests? They are testing both parse and stringify and current tests are split by this methods. To me it feels more reasonable to test them together in this case.

lib/parse.js
Coobaha: Could you provide input for this case? All previous inputs are still valid and i've covered this cases in tests:
```
&
&=
&=&
&=&=
&=&=&
=
=&
=&&&
=&=&=&
``` 

Do i miss something? Thanks

ljharb: I don't have any input off the top of my head :-) just looking at the code paths.

test/parse.js
```diff
@@ -843,3 +843,70 @@ test('parse()', function (t) {
...
+        { input: '%5B0%5D=a&%5B1%5D=b', withEmptyKeys: { '': ['a', 'b'] }, stringifyOutput: '[0]=a&[1]=b', noEmptyKeys: { 0: 'a', 1: 'b' } }
+    ];
+
+    testCases.forEach(function (testCase) {
+        t.test('parses an object with empty string key with ' + testCase.input, function (st) {
+            var parsed = qs.parse(testCase.input, { allowEmptyKeys: true });
+
+            st.deepEqual(parsed, testCase.withEmptyKeys);
+            st.deepEqual(qs.stringify(parsed, { encode: false }), testCase.stringifyOutput);
+            st.deepEqual(qs.parse(testCase.stringifyOutput, { allowEmptyKeys: true }), testCase.withEmptyKeys);
```
ljharb, line 888: i'd prefer to keep stringify tests in the stringify tests file; it's fine if the test cases are in a separate file both test files require.

Coobaha: Hey, sorry for the delay :) i've extracted test cases into own file and and separated stringify/parse parts accordingly

test/parse.js
```diff
@@ -843,3 +844,46 @@ test('parse()', function (t) {
...
+            { '': { '': ['2', '3'] } },
+            'array/object conversion',
+            { skip: 'TODO: figure out what this should do' }
+        );
+
+        st.deepEqual(
+            qs.parse('[][0]=2&[][1]=3&[a]=2', { allowEmptyKeys: true }),
+            { '': { '': ['2', '3'], a: '2' } },
+            'array/object conversion',
+            { skip: 'TODO: figure out what this should do' }
```
ljharb, line 884: let's discuss these TODOs. what are your thoughts?

test/stringify.js
```diff
@@ -861,3 +862,20 @@ test('stringify()', function (t) {
 
     t.end();
 });
+
+test('stringifies empty keys', function (t) {
+    emptyTestCases.forEach(function (testCase) {
+        t.test('stringifies an object with empty string key with ' + testCase.input, function (st) {
+            st.deepEqual(qs.stringify(testCase.withEmptyKeys, { encode: false }), testCase.stringifyOutput);
```
ljharb, line 869: can we also add a test case where the input is a stringified `noEmptyKeys`?

ljharb (commented):
Looking great!

test/parse.js
Coobaha: Since we parse empty keys I assume output should be equal to key-full strings? 
- If we use `a[b][0]=2&a[b][1]=3` the output object is `{ a: { b: [ '2', '3' ] } }`
- when we replace `a` and `b` with empty strings in this object: `{ '': { '': [ '2', '3' ] } }` and query string should be `[][0]=2&[][1]=3` which is `<EMPTY_KEY>[<EMPTY_KEY>][0]=2` 

Current output for `[][0]=2&[][1]=3` without empty keys flag is ` { 0: '2', 1: '3' }`

test/empty-keys-cases.js
```diff
@@ -2,36 +2,36 @@
...
-        { input: '&=', withEmptyKeys: { '': '' }, stringifyOutput: '=', noEmptyKeys: {} },
-        { input: '&=&', withEmptyKeys: { '': '' }, stringifyOutput: '=', noEmptyKeys: {} },
-        { input: '&=&=', withEmptyKeys: { '': ['', ''] }, stringifyOutput: '[0]=&[1]=', noEmptyKeys: {} },
-        { input: '&=&=&', withEmptyKeys: { '': ['', ''] }, stringifyOutput: '[0]=&[1]=', noEmptyKeys: {} },
+        { input: '&', withEmptyKeys: {}, stringifyOutput: '', noEmptyKeys: {}, stringifyOutputNoEmpty: '' },
+        { input: '&&', withEmptyKeys: {}, stringifyOutput: '', noEmptyKeys: {}, stringifyOutputNoEmpty: '' },
+        { input: '&=', withEmptyKeys: { '': '' }, stringifyOutput: '=', noEmptyKeys: {}, stringifyOutputNoEmpty: '' },
+        { input: '&=&', withEmptyKeys: { '': '' }, stringifyOutput: '=', noEmptyKeys: {}, stringifyOutputNoEmpty: '' },
+        { input: '&=&=', withEmptyKeys: { '': ['', ''] }, stringifyOutput: '[0]=&[1]=', noEmptyKeys: {}, stringifyOutputNoEmpty: '' },
+        { input: '&=&=&', withEmptyKeys: { '': ['', ''] }, stringifyOutput: '[0]=&[1]=', noEmptyKeys: {}, stringifyOutputNoEmpty: '' },
```
ljharb, line 10: it makes me *very* nervous to be modifying test cases for a bug fix. is there any chance for a default algorithm to infer `stringifyOutputNoEmpty` rather than needing to add it to each case?

Coobaha: 🤷 I think explicit cases are better here. I assume you thought that we should default to `''` if you want a smaller diff, then it also makes to do the same for `stringifyOutput`?

PR #435, opened 2022-03-11
新增keySort配置

ywzou:
对最终的key进行排序


ckcr4lyf:
Additionally, the keys are sorted prior to the mapping as well, so I would think this isn't necessary?

ljharb:
Closing, pending more context from the OP.

Issue #436, opened 2022-03-15
[invalid CVE]: “Property Injection” in the function merge() CVE-2021-44907


Marynk:
merge() (https://github.com/ljharb/qs/blob/main/dist/qs.js#L670) allows to assign properties on an array in the query. In case of any property being assigned a value the array is converted to an object containing these properties.
Essentially, this means that the property whose expected type is Array always has to be checked with `Array.isArray()` by the user, which may not be obvious to the user and can cause unexpected behavior. 
While this seems intentional, this behavior is not stressed in documentation. 

A couple of simple examples:
https://jsfiddle.net/1s7pq93z/1/
https://jsfiddle.net/65jxksay/

The CVE Program has assigned the ID CVE-2021-44907 to this issue. This is a record on the [CVE List](https://cve.mitre.org/cve/), which standardizes names for security problems.

ljharb:
I’m confused; why is there a CVE that wasn’t responsibly disclosed first?

ljharb:
None of these seem like vulnerabilities, after a deeper look. Prototype pollution attacks are about polluting *global prototypes*, not about polluting a single instance.

You're totally right that someone who is writing bad code won’t be properly checking for an array, and that someone writing non-robust code won’t be doing `Array.prototype.map.call`, but at best this is a usability improvement, not a security vulnerability.

We could certainly warn when an own property being set on an array is shadowing an Array.prototype method - that seems reasonable to me - but either way, i don’t consider any of these to be CVE-worthy. Most bugs aren’t vulnerabilities, they’re just bugs.

dougwilson:
I'm not sure if there is actually an issue with `Qs` here, like @ljharb suggested. In fact, it seems like the suggestion to add a warning when an object changes internally from array to object wouldn't actually "fix" the issue, as the Fiddle demonstrates that the input of `Qs.parse("arr.length=2&arr=a&arr=b&arr=c", {allowDots: true});` would result in the same output object, but there was no internal transition from an Array to an Object to warn about.

It seems there may not be any way to "fix" the issue without a validation framework, as the demo shows that if a user does not call `Array.isArray` on something they are expecting to be an array and proceeds to write a loop like `for (var i = 0; i < obj.list.length; i++) {}` the input can make it loop forever. This is equally trust with `JSON.parse` itself, as the user could provide an object in that position with a `length` key.

ljharb:
[...]

PR #440, opened 2022-04-07
[Fix] do not override array methods/properties (CVE-2021-44907)

lscoder:
This is a fix for [CVE-2021-44907](https://cve.mitre.org/cgi-bin/cvename.cgi?name=2021-44907#:~:text=A%20Denial%20of%20Service%20vulnerability,parse%20function.).
Any property from source object that already exists in `Array.prototype` is ignored.

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -68,11 +68,6 @@ var merge = function merge(target, source, options) {
         return [target].concat(source);
     }
 
-    var mergeTarget = target;
-    if (isArray(target) && !isArray(source)) {
-        mergeTarget = arrayToObject(target, options);
-    }
-
     if (isArray(target) && isArray(source)) {
         source.forEach(function (item, i) {
             if (has.call(target, i)) {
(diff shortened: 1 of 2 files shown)


ljharb:
See #436 - this CVE is INVALID. It is not a vulnerability, whatsoever.

ljharb (changes requested):
It should, and forever will, allow creating own properties that shadow Array.prototype ones, because that is not an attack vector.

ljharb:
In the future, please do not submit an unsolicited PR until you're sure the maintainer wants it. This PR's git ref is now permanent pollution on my repo.

lib/utils.js
```diff
@@ -89,8 +84,28 @@ var merge = function merge(target, source, options) {
         return target;
     }
 
-    return Object.keys(source).reduce(function (acc, key) {
-        var value = source[key];
+    var mergeTarget = target;
+    var mergeSource = source;
+
+    if (isArray(target) && !isArray(source)) {
+        var restrictedKeys = new Set(Object.getOwnPropertyNames(Array.prototype));
```
ljharb, line 91: this is also a breaking change; `qs` supports node 0.4, which does not have Set.

PR #447, opened 2022-07-13
[readme] add usage of `filter` option for injecting custom serialization, i.e. of custom types

gomain:
Add to README.md the usage of option `filter` for injecting custom serialization, i.e. of custom types.


ljharb (approved):
Thanks, seems like a great improvement

PR #463, opened 2023-01-20
Encode comma values more consistently

### LINKED ISSUE #459: `stringify` with `arrayFormat: 'comma'` does not encode commas in strings.
Hello, 

I noticed that calling `qs.stringify({ query: 'some, query' }, { arrayFormat: 'comma', }` would output `?query=some,query` which, if parsed back, would generate an array `{ query: ['some', 'query'] }`.

I looked at the code for `stringify` and I noticed this - https://github.com/ljharb/qs/blob/main/lib/stringify.js#L125

This will essentially split the string by the comma and then re-stitch it. I think this is used for handling nested objects with arrays in them? Not sure what would be the best way to fix this...

bakkerthehacker:
Fixes #459 but only for `encodeValuesOnly: true`

This PR includes tests for arrays with comma values and different settings.  

The behaviour for `encodeValuesOnly` has been adjusted to encode values before they are joined together, as an example it changes stringifying `{ a: [',', '', 'c,d'] }` from `a=,,,c,d` to `a=%2C,,c%2Cd`.  

The behaviour of the other 2 encoding setups (encode: false) and (encode: true with encodeValuesOnly: false) have been left unmodified.  These formats do seem less useful for `arrayFormat: 'comma'`, and tricky to get right in all situations.  

However, there is additional functionality in stringify that isnt really tested, but is related to these options for the types: 
'string' 'number' 'boolean' 'symbol' 'bigint' 'buffer'.  This inconsistent additional functionality also avoids encoding commas when they appear in these types.  

I have included additional tests for these situations, where a comma appears in a string outside an array.  I have also removed the custom functionality in this PR and these types now encode comma values normally.  If this change goes too far, it can be removed, but it seems odd to have the `comma` version of the arrayFormat affect non-array types just because they contain a comma.


coatesap (approved):
This looks great, and would definitely allow us to finally start using `arrayFormat` in production.

ljharb (commented):
This looks good overall, but there's a test failing - "array with multiple items with a comma inside". Does that mean this is a breaking change, or is it a bug in the PR?

ljharb:
Given that `qs.parse('a=c%25%2Cd%2Ce')` is `{ a: 'c%,d,e' }` and `qs.parse('a=c%2Cd%2Ce')` is `{ a: 'c,d,e' }`, this is certainly an improvement, but i think the ideal end point is where things can round trip.

I'll consider this semver-patch since for comma arrayFormat users it's clearly broken atm.

ljharb (approved):

PR #473, opened 2023-05-07
[Fix] `parse`: Fix parsing when the global Object prototype is frozen

jportner:
## Background

<details>
<summary>Click to expand</summary>

_Note: examples use the Node.js REPL with strict mode: `NODE_REPL_MODE=strict node`._

### Inheritance and Shadowing

Objects in JavaScript [inherit properties from their prototype chain](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Inheritance_and_the_prototype_chain). For example, the "toString" property can be accessed on all objects, but it doesn't actually exist on each object, it exists on the global Object prototype:

```js
let obj = {};
obj.toString();
// '[object Object]'
Object.getOwnPropertyDescriptor(obj, 'toString');
// undefined
Object.getOwnPropertyDescriptor(Object.prototype, 'toString');
// { value: [Function: toString], writable: true, enumerable: false, configurable: true }
```

Under normal circumstances, you can assign a property to an object using the `=` operator, and any property of the same name in the object's prototype chain will not be modified, but will be "[shadowed](https://en.wikipedia.org/wiki/Variable_shadowing)" by the new property:

```js
obj.toString = () => 'foo';
obj.toString();
// 'foo'
Object.getOwnPropertyDescriptor(obj, 'toString');
// { value: [Function: toString], writable: true, enumerable: true, configurable: true }
```

### Prototype Pollution

[From Snyk:](https://learn.snyk.io/lessons/prototype-pollution/javascript/)

> Prototype pollution is an injection attack that targets JavaScript runtimes. With prototype pollution, an attacker might control the default values of an object's properties. This allows the attacker to tamper with the logic of the application and can also lead to denial of service or, in extreme cases, remote code execution.

There are a few different ways to mitigate Prototype Pollution, and one way to do it across the board is to [freeze the global "root" objects and their prototypes (Object, Function, Array, etc.)](https://www.npmjs.com/package/nopp)

[From MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/freeze):

> The `Object.freeze()` static method *freezes* an object. Freezing an object [prevents extensions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/preventExtensions) and makes existing properties non-writable and non-configurable. A frozen object can no longer be changed: new properties cannot be added, existing properties cannot be removed, their enumerability, configurability, writability, or value cannot be changed, and the object's prototype cannot be re-assigned.

This means that any attempt to change the Object prototype will fail. If using [strict mode](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode), it will throw an error; otherwise, it will be silently ignored.

If the Object prototype becomes frozen, all of its properties are no longer writable or configurable:

```js
Object.freeze(Object.prototype);
Object.getOwnPropertyDescriptor(Object.prototype, 'toString');
// { value: [Function: toString], writable: false, enumerable: false, configurable: false }
```

This also prevents shadowing properties with assignment. If an object _doesn't_ already have a property defined (such as "toString"), **and** it inherits a _non-writable_ property of that name from its prototype chain, any attempt to assign the property on that object will fail:

```js
let obj2 = {};
obj2.toString = () => 'bar';
// Uncaught TypeError: Cannot assign to read only property 'toString' of object '#<Object>'
obj2.toString();
// '[object Object]'
```

This behavior is described in the [ECMAScript 2016 specification](https://262.ecma-international.org/7.0/#sec-strict-mode-of-ecmascript):

> Assignment to an undeclared identifier or otherwise unresolvable reference does not create a property in the [global object](https://262.ecma-international.org/7.0/#global-object). When a simple assignment occurs within [strict mode code](https://262.ecma-international.org/7.0/#sec-strict-mode-code), its [LeftHandSideExpression](https://262.ecma-international.org/7.0/#prod-LeftHandSideExpression) must not evaluate to an unresolvable [Reference](https://262.ecma-international.org/7.0/#sec-reference-specification-type). If it does a **ReferenceError** exception is thrown ([6.2.3.2](https://262.ecma-international.org/7.0/#sec-putvalue)). The [LeftHandSideExpression](https://262.ecma-international.org/7.0/#prod-LeftHandSideExpression) also may not be a reference to a data property with the attribute value {[[Writable]]: **false**}, to an accessor property with the attribute value {[[Set]]: **undefined**}, nor to a non-existent property of an object whose [[Extensible]] internal slot has the value **false**. In these cases a `TypeError` exception is thrown ([12.15](https://262.ecma-international.org/7.0/#sec-assignment-operators)).

</details>

## The Problem

The `allowPrototypes` option is designed to strip any query parameters if they would shadow/overwrite properties on the global Object prototype (for example, "toString"). This logic is in the `parseKeys()` function:

https://github.com/ljharb/qs/blob/9dca37f15de317fe9ad0ced907cdf250ba310880/lib/parse.js#L172-L177

However, before calling `parseKeys()`, the parse module first calls `parseValues()` to create a temporary object:

https://github.com/ljharb/qs/blob/9dca37f15de317fe9ad0ced907cdf250ba310880/lib/parse.js#L246-L256

Unfortunately, when you have frozen the global Object prototype, the `parseValues()` function will throw a TypeError when inherited properties are present. Example:

```js
Object.freeze(Object.prototype);
const { parse } = require('./lib');
parse('foo=bar&toString=123');
// Uncaught TypeError: Cannot assign to read only property 'toString' of object '#<Object>'
```

This means that projects cannot use this package if they have frozen the global Object prototype.

## The Solution

Since the `parseValues()` function returns a temporary object, the simple solution here is to create a "plain object" that does not have a prototype. This doesn't require any other logic changes, because the if the `allowPrototypes` option is set to false, the `parseKeys()` function will subsequently strip out any problematic properties.

With this change, this package will be compatible with the "frozen prototype" approach of mitigating Prototype Pollution 🎉

Now you can use the same example as above, and it works as expected:

```js
Object.freeze(Object.prototype);
const { parse } = require('./lib');
parse('foo=bar&toString=123');
// { foo: 'bar' }
```


dist/qs.js
```diff
@@ -323,7 +329,6 @@ var arrayPrefixGenerators = {
 };
 
 var isArray = Array.isArray;
-var split = String.prototype.split;
```
jportner, line 326: I ran `npm run dist` to rebuild the browserified distributable, and it made a bunch of changes unrelated to my PR.

It looks like maybe you forgot to run this before the 6.11.1 release?

ljharb: Nope, it shouldn't *be* run except as part of a release, so the changes to this file should be reverted.

joe-matsec: Ah okay, thanks, done in 8dc6db802018ef7c05d4bcf63b28b38723677da2

jportner: Ah okay thanks, done in 8dc6db802018ef7c05d4bcf63b28b38723677da2

lib/parse.js
```diff
@@ -49,7 +49,13 @@ var isoSentinel = 'utf8=%26%2310003%3B'; // encodeURIComponent('&#10003;')
 var charsetSentinel = 'utf8=%E2%9C%93'; // encodeURIComponent('✓')
 
 var parseValues = function parseQueryStringValues(str, options) {
-    var obj = {};
+    // This creates a temporary object to hold parsed query string values.
+    // We need to use a plain object in case the global Object prototype is frozen, otherwise we could encounter a TypeError here if the
+    // query string contains properties that exist on the Object prototype.
+    // If the "allowPrototypes" option is set to false (default), then any query properties that _would_ overwrite properties on the Object
+    // prototype will be stripped by the parseKeys function before being returned.
+    var obj = Object.create(null);
```
ljharb, line 57:
```suggestion
    var obj = { __proto__: null };
```
instead of the comments, how about a test?

ljharb (changes requested):
We'd need some tests to cover this.

One challenge, though, is that this would fail in an env where Object.create isn't available (if Object.prototype is frozen, violating the spec, then Object.create surely might be deleted), which hasn't been the case by default before.

However, the suggestion below would address that. Either way, we'd need tests to cover it.

jportner:
> Unfortunately, when you have frozen the global Object prototype, the `parseKeys()` function will throw a TypeError when inherited properties are present. Example:

Whoops, I wrote `parseKeys()` here when I meant to write `parseValues()`, that is what was throwing the error. I edited the PR description accordingly.

ljharb:
Note that `node --use_strict` is NOT "strict mode", is something v8-specific, and should never, ever be used. The only proper strict mode is with `'use strict';` or in a class body, or an ES Module.

lib/parse.js
jportner: Sure, I can add a test!

Regarding this comment:

> (if Object.prototype is frozen, violating the spec, then Object.create surely might be deleted)

I'm happy to change it. If this is a concern, do you think other existing usages of `Object.create(null)` should be updated as well?

ljharb: Yes, but let's not do that in this PR. I can do it separately.

jportner:
> Note that `node --use_strict` is NOT "strict mode", is something v8-specific, and should never, ever be used. The only proper strict mode is with `'use strict';` or in a class body, or an ES Module.

TIL, thanks for pointing this out!

Apparently the correct way to enable strict mode in the Node.js REPL is to use the [`NODE_REPL_MODE=strict` env var](https://nodejs.org/docs/latest-v16.x/api/repl.html#environment-variable-options).

lib/parse.js
jportner: Okay, I added a test in 64d88946b0943e908e86e52c41c739a43418ffc2, ready for another pass.

I had to get a little creative, but I think it's sufficient. Let me know what you think, thanks!

ljharb (approved):
Refactored a bit to use `mock-property` and leave it configurable, since that isn't relevant to the test

ljharb:
node < 6 has a different error message; i fixed that locally with a regex, but node <= 0.8 actually doesn't throw when Object.prototype is frozen, so I'll make a package to detect that, and update this, and then land it.

joe-matsec:
Thank you!

PR #479, opened 2023-07-28
[New] `parse`: call decoder when parsing objects

### LINKED ISSUE #284: Decoder function not being called
Hi,
I'm getting crazy with this but for some reason the decoder function seems to never be called.
Here is my code:
```typescript
  router.get('/testqs', function(req, res) {
    const result = qs.parse(req.query, {
      decoder: function (str) {
        let result: any
        const keywords: any = {
          true: true,
          false: false,
          null: null,
          undefined: undefined
        }

        if (str in keywords) {
          result = keywords[str]
        } else {
          const parsed = parseFloat(str)
          if (isNaN(parsed)) {
            result = str
          } else {
            result = parsed
          }
        }

        console.log(str, result)
        return result
      }
    })
    res.json(result)
  })
```

When I send a request:
```bash
GET {{host}}{{api}}/testqs?maintYear=2018&maintMonth=5&fields[]=id&fields[]=regname&fields[]=price&notRenewed=true
```

I get the following result:

```json
{
  "maintYear": "2018",
  "maintMonth": "5",
  "fields": [
    "id",
    "regname",
    "price"
  ],
  "notRenewed": "true"
}
```

To summarize, evrrything seems to work properly except that the decoder function seems to never be called (and of course, I don't have any console.log output from inside the function).

I noticed that someone in the following comment (https://github.com/ljharb/qs/issues/91#issuecomment-358708095) that someone is saying that this doesn't seem to work for him but not sure if he doesn't like the result or if he encounters the same problem as I do.

Any idea regarding what could go wrong ?

Thanks

nbgraham:
Fixes #284.

### DIFF
--- a/lib/parse.js
+++ b/lib/parse.js
@@ -238,13 +238,10 @@ var normalizeParseOptions = function normalizeParseOptions(opts) {
 };
 
 var decodeObject = function (obj, opts) {
-    if (opts.decoder === defaults.decoder) {
-        return obj;
-    }
     if (typeof obj !== 'object') {
         return obj;
     }
-    var decodedObj = {};
+    var decodedObj = opts.plainObjects ? Object.create(null) : {};
     var keys = Object.keys(obj);
     for (var i = 0; i < keys.length; ++i) {
         var key = keys[i];
--- a/lib/parse.js
+++ b/lib/parse.js
@@ -237,6 +237,28 @@ var normalizeParseOptions = function normalizeParseOptions(opts) {
     };
 };
 
+var decodeObject = function (obj, opts) {
+    if (opts.decoder === defaults.decoder) {
+        return obj;
+    }
+    if (typeof obj !== 'object') {
+        return obj;
+    }
+    var decodedObj = {};
+    var keys = Object.keys(obj);
+    for (var i = 0; i < keys.length; ++i) {
+        var key = keys[i];
+        var val = obj[key];
+        var decodedKey = opts.decoder(key, defaults.decoder, opts.charset, 'key');
+        var decodedVal = opts.decoder(val, defaults.decoder, opts.charset, 'value');
+        decodedVal = utils.maybeMap(decodedVal, function (mapped) {
+            return decodeObject(mapped, opts);
+        });
+        decodedObj[decodedKey] = decodedVal;
+    }
+    return decodedObj;
+};
+
 module.exports = function (str, opts) {
     var options = normalizeParseOptions(opts);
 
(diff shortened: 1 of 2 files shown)


lib/parse.js
```diff
@@ -237,14 +237,36 @@ var normalizeParseOptions = function normalizeParseOptions(opts) {
     };
 };
 
+var decodeObject = function (obj, opts) {
+    if (opts.decoder === defaults.decoder) {
+        return obj;
+    }
```
ljharb, line 243: i think this optimization is actually incorrect, because the default decoder does actual work.
```suggestion
```

lib/parse.js
```diff
@@ -237,14 +237,36 @@ var normalizeParseOptions = function normalizeParseOptions(opts) {
...
+    if (typeof obj !== 'object') {
+        return obj;
+    }
+    var decodedObj = {};
```
ljharb, line 247: this needs to respect `plainObjects`.
```suggestion
    var decodedObj = opts.plainObjects ? Object.create(null) : {};
```

test/parse.js
```diff
@@ -767,6 +767,77 @@ test('parse()', function (t) {
...
+            { val: true, type: 'value' }
+        ]);
+        st.end();
+    });
+
+    t.test('can parse object with custom encoding', function (st) {
+        var keywords = {
+            'true': true
+        };
+        st.deepEqual(qs.parse({
```
ljharb, line 804:
```suggestion
        st.deepEqual(
            qs.parse({
```

test/parse.js
```diff
@@ -767,6 +767,77 @@ test('parse()', function (t) {
...
+            maintYear: '2018',
+            maintMonth: '5',
+            fields: [
+                'id',
+                'regname',
+                'price'
+            ],
+            notRenewed: 'true'
+        }, {
```
ljharb, line 813:
```suggestion
        },
        {
```

test/parse.js
```diff
@@ -767,6 +767,77 @@ test('parse()', function (t) {
...
+                    var parsed = parseFloat(str);
+                    if (isNaN(parsed)) {
+                        return str;
+                    }
+                    return parsed;
+
+                }
+                return str;
+            }
+        }), {
```
ljharb, line 828:
```suggestion
        }),
        {
```

ljharb (changes requested):

ljharb:
This seems like it's either a bugfix - a patch - or a breaking change, in which case it should be implemented behind an option instead.

nbgraham:
@ljharb 
Based on [this](https://github.com/ljharb/qs/issues/284#issuecomment-426098167) I was treating it as a bugfix. It's kinda already behind the `decoder` option. If you don't provide that, this won't do anything. 
But I guess someone could be providing the `decoder` option today, and this could change the behavior for them.

ljharb:
That's exactly my concern - especially since this issue was filed *five* years ago and the behavior probably long predates that.

ljharb:
@nbgraham any interest in completing this PR?

nbgraham:
> @nbgraham any interest in completing this PR?

No, it doesn't have a clear solution and it's not worth it to me

PR #484, opened 2023-11-03
Add qs logos, use in README

agreea:
Adds the QS logo to the README and references [NUMI Open Design](https://github.com/numi-hq/open-design).


README.md
```diff
@@ -1,3 +1,7 @@
+<p align="center">
+  <img alt="qs" src="./logos/banner_dark.png" width="800" />
```
ljharb, line 2: hm, the banner ends up pretty tall here (https://github.com/agreea/qs/tree/add-logo#readme) - could we make something that's a lot shorter so it doesn't push the readme content so far down the fold?

ljharb (approved):
would be good to have one that's even half the height, but let's go ahead with this :-)

PR #488, opened 2024-01-28
[New] `stringify`/`parse`: add `encodeDotKeys`/`decodeDotKeys` respectively to allow . in object keys

### LINKED ISSUE #249: If { allowDots: true } is passed, dots in property names should be encoded
Currently `Qs.stringify({ 'this.is.unfortunate': 2}, { allowDots: true })` results in `'this.is.unfortunate=2'`, which is ambiguous. `'this.is.unfortunate=2'` can be interpreted as both `{ this: { is: { unfortunate: 2 } }`, and also `{ 'this.is': { unfortunate: 2 } }`, and also `{ this: { 'is.unfortunate': 2 } }`. Truly this is unfortunate - see my fiddle [here](https://jsfiddle.net/z95qox6j/1/).

I propose that Qs.stringify() encode keys that dots are escaped and Qs.parse() decode keys in the reverse manner. The goal would be to make Qs.parse(Qs.stringify(x)) lossless - at least in this particular case.

Open Questions:
1. Is this a good idea?
2. What encoding should be used?
3. Should encoding always occur, or only when { allowDots: true } is passed?

Thank you!

aks-:
Previously the dot in the keys would be encoded but the round-tripping was broken. Here is the previous behavior
```javascript
const opts = { allowDots: true };
qs.parse(qs.stringify({ "name.obj": { "first": "John", "last": "Doe" } }, opts), opts)
 // => { name: { obj: { first: 'John', last: 'Doe' } } }
 ```
 
This PR fixes the above behavior and fixes #249:
```javascript
 qs.parse(qs.stringify({ "name.obj": { "first": "John", "last": "Doe" } }, { allowDots: true, encodeDotKeys: true }), { allowDots: true, decodeDotKeys: true })
{ 'name.obj': { first: 'John', last: 'Doe' } }
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,6 +151,14 @@ var withDots = qs.parse('a.b=c', { allowDots: true });
 assert.deepEqual(withDots, { a: { b: 'c' } });
 ```
 
+Option `decodeDotInKeys` can be used to decode dots in keys
+Note: it implies `allowDots`, so `parse` will error if you set `decodeDotInKeys` to `true`, and `allowDots` to `false`.
+
+```javascript
+var withDots = qs.parse('name%252Eobj.first=John&name%252Eobj.last=Doe', { decodeDotInKeys: true });
+assert.deepEqual(withDots, { 'name.obj': { first: 'John', last: 'Doe' }});
+```
+
 Option `allowEmptyArrays` can be used to allowing empty array values in object
 ```javascript
 var withEmptyArrays = qs.parse('foo[]&bar=baz', { allowEmptyArrays: true });
@@ -424,6 +432,14 @@ You may override this to use dot notation by setting the `allowDots` option to `
 ```javascript
 qs.stringify({ a: { b: { c: 'd', e: 'f' } } }, { allowDots: true });
 // 'a.b.c=d&a.b.e=f'
+```
+
+You may encode the dot notation in the keys of object with option `encodeDotInKeys` by setting it to `true`:
+Note: it implies `allowDots`, so `stringify` will error if you set `decodeDotInKeys` to `true`, and `allowDots` to `false`.
+Caveat: when `encodeValuesOnly` is `true` as well as `encodeDotInKeys`, only dots in keys and nothing else will be encoded.
+```javascript
+qs.stringify({ "name.obj": { "first": "John", "last": "Doe" } }, { allowDots: true, encodeDotInKeys: true })
+// 'name%252Eobj.first=John&name%252Eobj.last=Doe'
 ```
 
 You may allow empty array values by setting the `allowEmptyArrays` option to `true`:
--- a/lib/parse.js
+++ b/lib/parse.js
@@ -229,8 +229,10 @@ var normalizeParseOptions = function normalizeParseOptions(opts) {
         throw new TypeError('The duplicates option must be either combine, first, or last');
     }
 
+    var allowDots = typeof opts.allowDots === 'undefined' ? defaults.allowDots : !!opts.allowDots;
+
     return {
-        allowDots: typeof opts.allowDots === 'undefined' ? defaults.allowDots : !!opts.allowDots,
+        allowDots: allowDots,
         allowEmptyArrays: typeof opts.allowEmptyArrays === 'boolean' ? !!opts.allowEmptyArrays : defaults.allowEmptyArrays,
         allowPrototypes: typeof opts.allowPrototypes === 'boolean' ? opts.allowPrototypes : defaults.allowPrototypes,
         allowSparse: typeof opts.allowSparse === 'boolean' ? opts.allowSparse : defaults.allowSparse,
--- a/lib/parse.js
+++ b/lib/parse.js
@@ -14,6 +14,7 @@ var defaults = {
     charset: 'utf-8',
     charsetSentinel: false,
     comma: false,
+    decodeDotInKeys: true,
     decoder: utils.decode,
     delimiter: '&',
     depth: 5,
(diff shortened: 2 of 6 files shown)


aks-:
@ljharb I wonder if we need `allowDots` to be `true` for `encodeDotKeys` and `decodeDotKeys` work. Also I can not think of any other name that is similar for `parse` and `stringify` method. Any suggestions?

lib/parse.js
```diff
@@ -119,7 +120,7 @@ var parseObject = function (chain, val, options, valuesParsed) {
 
     for (var i = chain.length - 1; i >= 0; --i) {
         var obj;
-        var root = chain[i];
+        var root = options.allowDots && options.decodeDotKeys ? chain[i].replace(/%2E/g, '.') : chain[i];
```
ljharb, line 123: please add a test case with more than one dot in the key

ljharb (commented):
Please add test cases that use encodeValuesOnly with this option.

README.md
```diff
@@ -149,6 +149,14 @@ var withDots = qs.parse('a.b=c', { allowDots: true });
 assert.deepEqual(withDots, { a: { b: 'c' } });
 ```
 
+Option `decodeDotInKeys` can be used to decode dot notations in keys
+Note: it only works with `allowDots` so you have to provide both options to be `true`:
+
+```javascript
+var withDots = qs.parse('name%252Eobj.first=John&name%252Eobj.last=Doe', { allowDots: true, decodeDotInKeys: true });
+assert.deepEqual(withDots, { 'name.obj': { first: 'John', last: 'Doe' }});
+```
```
ljharb, line 158: now that i think about it, do we need a parse option here? i feel like it should just be automatically decoded. what tests break if we default this option to `true`?

test/parse.js
```diff
@@ -76,6 +76,88 @@ test('parse()', function (t) {
...
+        st.deepEqual(
+            qs.parse('name%252Eobj.first=John&name%252Eobj.last=Doe', { allowDots: true, decodeDotInKeys: false }),
+            { 'name%2Eobj': { first: 'John', last: 'Doe' } },
+            'with allowDots true and decodeDotInKeys false'
+        );
+        st.deepEqual(
+            qs.parse('name%252Eobj.first=John&name%252Eobj.last=Doe', { allowDots: true, decodeDotInKeys: true }),
+            { 'name.obj': { first: 'John', last: 'Doe' } },
+            'with allowDots true and decodeDotInKeys true'
+        );
```
ljharb, line 100: these are all 4 of the combinations of these two options. the problem is that there's only 3 variations here.

We need to figure out if having 4 variations makes sense, and what that behavior would be - and if we can't come up with anything, then that 4th combination needs to throw.

test/stringify.js
```diff
@@ -64,6 +64,116 @@ test('stringify()', function (t) {
...
+            'with allowDots false and encodeDotInKeys true'
+        );
+        st.equal(
+            qs.stringify(
+                { 'name.obj': { first: 'John', last: 'Doe' } },
+                { allowDots: true, encodeDotInKeys: true }
+            ),
+            'name%252Eobj.first=John&name%252Eobj.last=Doe',
+            'with allowDots true and encodeDotInKeys true'
+        );
```
ljharb, line 99: same comment here

ljharb (changes requested):

test/parse.js
aks-: I think the second combination should throw here `with allowDots false and decodeDotInKeys true` because it ideally means decoding to invalid state like `{ name.obj.first: 'John', name.obj.last: 'Doe' }` which makes it very ambiguous. What do you think?

test/stringify.js
aks-: Here second one as well. It does not do correct round tripping, I will push a commit

test/parse.js
ljharb: the point of `allowDots` is to treat a dot in a key as a nested object marker. when false, `name.obj.first` would just be either literally `name.obj.first` or `name%2Eobj%2Efirst`, no? what about that is invalid?

aks-: @ljharb, the problem is that round-tripping breaks here. Let's consider the following example: `qs.stringify({ 'name.obj': { first: 'John', last: 'Doe' } }, { allowDots: true, encodeDotInKeys: false })`. Now, the counter `options` for the string generated by this would be `{ allowDots: true, decodeDotInKeys: false }`, and the consumer might think that they would get the object they provided to `stringify`. However, the parser cannot distinguish between a key and a dot notation for key/value separation in `'name.obj.first=John&name.obj.last=Doe'`.

```javascript
qs.parse(qs.stringify({ 'name.obj': { first: 'John', last: 'Doe' } }, { allowDots: true, encodeDotInKeys: false }),{ allowDots: true, decodeDotInKeys: false } )
// { name: { obj: { first: 'John', last: 'Doe' } } }
```

So, I believe both combinations (one for stringify and one for parse) should be considered illegal, right? One could argue that if `stringify` already throws for the combination, then the string would never be generated, but we never know how the consumer tends to use the package. What do you think? 🤔

ljharb: In that case, should `encodeDotInKeys` *imply* `allowDots: true` when allowDots is not passed? it'd still throw when allowDots is false.

aks-: Hm, I see why current behaviour works but is confusing. As a consumer I would expect `{ encodeDotInKeys: true }` to encode the keys. But currently we do not with this line `var encodedKey = allowDots && encodeDotInKeys ? key.replace(/\./g, '%2E') : key;`. I will fix it and push a change.

ljharb (approved):
Thanks! This looks good; I can clean up the remaining comments.

aks-:
@ljharb I realised that `allowDots = false & encodeDotInKeys` does not break the round-tripping but it does not `encode` the dot in key. I think this can be confusing so I will see if I fix that what other tests break and take it from there. I will keep you updated on this.

ljharb (approved):

PR #502, opened 2024-04-11
[Fix] utils: Optimize performance under large data volumes, reduce me…

lixinGiting:
When using qs.stringify to handle large data on a Windows Node.js environment resulted OOM. The problem was identified to occur during the process of concatenating a large number of strings, which consumed significant memory and time.

To resolve this issue, implemented a chunk-based processing approach. 

Tested this solution using 20MB of Chinese characters and observed the following improvements:
Memory usage was reduced by approximately 80%, from 1637MB to 327MB.
Processing speed increased by approximately 4 times, from 5256ms to 1318ms.

## Summary by CodeRabbit

- **Refactor**
	- Improved the efficiency of the string encoding process by handling input in segments, enhancing Unicode character support.

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -122,6 +122,10 @@ var decode = function (str, decoder, charset) {
     }
 };
 
+var limit = 1024;
+
+/* eslint operator-linebreak: [2, "before"] */
+
 var encode = function encode(str, defaultEncoder, charset, kind, format) {
     // This code was originally written by Brian White (mscdex) for the io.js core querystring library.
     // It has been adapted here for stricter adherence to RFC 3986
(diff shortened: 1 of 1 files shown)


lib/utils.js
```diff
@@ -143,45 +143,51 @@ var encode = function encode(str, defaultEncoder, charset, kind, format) {
...
-            || (c >= 0x30 && c <= 0x39) // 0-9
-            || (c >= 0x41 && c <= 0x5A) // a-z
-            || (c >= 0x61 && c <= 0x7A) // A-Z
-            || (format === formats.RFC1738 && (c === 0x28 || c === 0x29)) // ( )
-        ) {
-            out += string.charAt(i);
-            continue;
-        }
+    for (var j = 0; j < string.length; j += 1204) {
+        var segment = string.slice(j, j + 1204);
```
ljharb, line 147:
```suggestion
        var segment = string.length >= 1204 ? string.slice(j, j + 1204) : string;
```

lib/utils.js
```diff
@@ -143,45 +143,51 @@ var encode = function encode(str, defaultEncoder, charset, kind, format) {
...
-            || c === 0x7E // ~
-            || (c >= 0x30 && c <= 0x39) // 0-9
-            || (c >= 0x41 && c <= 0x5A) // a-z
-            || (c >= 0x61 && c <= 0x7A) // A-Z
-            || (format === formats.RFC1738 && (c === 0x28 || c === 0x29)) // ( )
-        ) {
-            out += string.charAt(i);
-            continue;
-        }
+    for (var j = 0; j < string.length; j += 1204) {
```
ljharb, line 146: let's store this `1204` in a constant. also, why was this number chosen?

ljharb (commented):
this PR adds a dependency on string slice, and array join (`arr.push(item)` can be converted to `arr[arr.length] = item`), which is undesirable, but otherwise it seems like a reasonable and straightforward change.

lib/utils.js
lixinGiting: Great suggestion. As for why 1204 was chosen, it's based on multiple experiments found that using 1024 as the split method resulted in the fastest speed and least memory usage.

ljharb: 1204 !== 1024 tho

lixinGiting: sry, 1024 is the correct value.  This is a foolish mistake on my part

lixinGiting: string.slice and array.join is undesirable. Is it due to compatibility considerations?

ljharb: Robustness - it's more methods that could be modified or deleted in the environment beforehand.

That's not a blocker here, it's just unfortunate.

lixinGiting:
I have made the changes according to your suggestions. Could you please review it again?

ljharb (approved):

lixinGiting:
Thanks a lot.
Codecov seems to be experiencing some issues, pending for several hours.

PR #511, opened 2024-08-01
[New] add strictDepth option to throw when input is beyond depth

jonchurch:
This PR proposes an option (`strictDepth`) to enforce throwing when input depth is beyond the `depth` option.

Defaults to `false`

Throws RangeError.

if `depth` has been set by the user to 0, we do not throw, but fallback to the previous behavior.

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,7 +115,18 @@ var deep = qs.parse('a[b][c][d][e][f][g][h][i]=j', { depth: 1 });
 assert.deepEqual(deep, { a: { b: { '[c][d][e][f][g][h][i]': 'j' } } });
 ```
 
-The depth limit helps mitigate abuse when **qs** is used to parse user input, and it is recommended to keep it a reasonably small number.
+You can configure **qs** to throw an error when parsing nested input beyond this depth using the `strictDepth` option (defaulted to false):
+
+```javascript
+try {
+    qs.parse('a[b][c][d][e][f][g][h][i]=j', { depth: 1, strictDepth: true });
+} catch (err) {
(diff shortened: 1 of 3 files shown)


wesleytodd (approved):

ljharb (approved):
mfw i see an extremely rare PR that requires zero tweaks, rebasing, commit message tweaking etc and i can just ff merge immediately

README.md
```diff
@@ -115,7 +115,18 @@ var deep = qs.parse('a[b][c][d][e][f][g][h][i]=j', { depth: 1 });
...
+```javascript
+try {
+    qs.parse('a[b][c][d][e][f][g][h][i]=j', { depth: 1, strictDepth: true });
+} catch (err) {
+    assert(err instanceof RangeError);
+    assert.strictEqual(err.message, 'Input depth exceeded depth option of 1 and strictDepth is true');
+}
+```
+
+The depth limit helps mitigate abuse when **qs** is used to parse user input, and it is recommended to keep it a reasonably small number. The strictDepth option adds a layer of protection by throwing an error when the limit is exceeded, allowing you to catch handle such cases.
```
jonchurch, line 129:
```suggestion
The depth limit helps mitigate abuse when **qs** is used to parse user input, and it is recommended to keep it a reasonably small number. The strictDepth option adds a layer of protection by throwing an error when the limit is exceeded, allowing you to catch and handle such cases.
```

jonchurch:
pushing a typo fix in readme

README.md
jonchurch: fixed and force pushed

Issue #515, opened 2024-10-27
`parse`: add new option to throw when `parameterLimit` is reached


ljharb:
Perhaps instead, add a new option that throws whenever *any* limit is reached, including `arrayLimit`.

See https://github.com/expressjs/express/issues/5878

IamLizu:
Hey @ljharb 👋 

I would like to work on this.

ljharb:
@IamLizu no need to "claim" issues in open source :-) just make a PR for anything that has a "help wanted" label

IamLizu:
Hey @ljharb 👋 

Can we do the implement the `arrayLimit` check in a different PR? If yes, I would enable #517 for review.

ljharb:
I think it'd be important to add it in the same PR, otherwise we'd need two options.

IamLizu:
Agreed 🤝

And thanks for the early review on the PR 🙏

PR #517, opened 2024-11-12
[New]: `parse`: add `throwOnLimitExceeded` option

IamLizu:
As discussed in https://github.com/expressjs/express/issues/5878 about adding an option to enable throwing error when parameter limit exceeded, I have, 

- [x] added `throwOnLimitExceeded` which is `false` by default.
- [x] added test cases to cover this change.
- [x] updated README to document this behavior.

Additionally, as https://github.com/ljharb/qs/issues/515 suggests about throwing an error in case of any limit is exceeded, it has been implemented as well.

#### References
1. Fixes https://github.com/ljharb/qs/issues/515

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,6 +135,18 @@ var limited = qs.parse('a=b&c=d', { parameterLimit: 1 });
 assert.deepEqual(limited, { a: 'b' });
 ```
 
+If you want an error to be thrown whenever the a limit is exceeded (eg, `parameterLimit`, `arrayLimit`), set the `throwOnLimitExceeded` option to `true`. This option will generate a descriptive error if the query string exceeds a configured limit.
+```javascript
+try {
+    qs.parse('a=1&b=2&c=3&d=4', { parameterLimit: 3, throwOnLimitExceeded: true });
+} catch (err) {
+    assert(err instanceof Error);
+    assert.strictEqual(err.message, 'Parameter limit exceeded. Only 3 parameters allowed.');
+}
+```
+
+When `throwOnLimitExceeded` is set to `false` (default), **qs** will parse up to the specified `parameterLimit` and ignore the rest without throwing an error.
+
 To bypass the leading question mark, use `ignoreQueryPrefix`:
 
 ```javascript
@@ -285,6 +297,18 @@ This limit can be overridden by passing an `arrayLimit` option:
 var withArrayLimit = qs.parse('a[1]=b', { arrayLimit: 0 });
 assert.deepEqual(withArrayLimit, { a: { '1': 'b' } });
 ```
+
+If you want to throw an error whenever the array limit is exceeded, set the `throwOnLimitExceeded` option to `true`. This option will generate a descriptive error if the query string exceeds a configured limit.
+```javascript
+try {
+    qs.parse('a[1]=b', { arrayLimit: 0, throwOnLimitExceeded: true });
+} catch (err) {
+    assert(err instanceof Error);
+    assert.strictEqual(err.message, 'Array limit exceeded. Only 0 elements allowed in an array.');
+}
+```
+
+When `throwOnLimitExceeded` is set to `false` (default), **qs** will parse up to the specified `arrayLimit` and if the limit is exceeded, the array will instead be converted to an object with the index as the key
 
 To disable array parsing entirely, set `parseArrays` to `false`.
 
--- a/lib/parse.js
+++ b/lib/parse.js
@@ -34,9 +34,13 @@ var interpretNumericEntities = function (str) {
     });
 };
 
-var parseArrayValue = function (val, options) {
+var parseArrayValue = function (val, options, currentArrayLength) {
     if (val && typeof val === 'string' && options.comma && val.indexOf(',') > -1) {
         return val.split(',');
+    }
+
+    if (options.throwOnLimitExceeded && currentArrayLength >= options.arrayLimit) {
+        throw new RangeError('Array limit exceeded. Only ' + options.arrayLimit + ' element' + (options.arrayLimit === 1 ? '' : 's') + ' allowed in an array.');
     }
 
     return val;
@@ -57,8 +61,17 @@ var parseValues = function parseQueryStringValues(str, options) {
 
     var cleanStr = options.ignoreQueryPrefix ? str.replace(/^\?/, '') : str;
     cleanStr = cleanStr.replace(/%5B/gi, '[').replace(/%5D/gi, ']');
+
     var limit = options.parameterLimit === Infinity ? undefined : options.parameterLimit;
-    var parts = cleanStr.split(options.delimiter, limit);
+    var parts = cleanStr.split(
+        options.delimiter,
+        options.throwOnLimitExceeded ? limit + 1 : limit
+    );
+
+    if (options.throwOnLimitExceeded && parts.length > limit) {
+        throw new RangeError('Parameter limit exceeded. Only ' + limit + ' parameter' + (limit === 1 ? '' : 's') + ' allowed.');
+    }
+
     var skipIndex = -1; // Keep track of where the utf8 sentinel was found
     var i;
 
@@ -93,8 +106,13 @@ var parseValues = function parseQueryStringValues(str, options) {
             val = options.strictNullHandling ? null : '';
         } else {
             key = options.decoder(part.slice(0, pos), defaults.decoder, charset, 'key');
+
             val = utils.maybeMap(
-                parseArrayValue(part.slice(pos + 1), options),
+                parseArrayValue(
+                    part.slice(pos + 1),
+                    options,
+                    isArray(obj[key]) ? obj[key].length : 0
+                ),
                 function (encodedVal) {
                     return options.decoder(encodedVal, defaults.decoder, charset, 'value');
                 }
@@ -121,7 +139,13 @@ var parseValues = function parseQueryStringValues(str, options) {
 };
 
 var parseObject = function (chain, val, options, valuesParsed) {
-    var leaf = valuesParsed ? val : parseArrayValue(val, options);
+    var currentArrayLength = 0;
+    if (chain.length > 0 && chain[chain.length - 1] === '[]') {
+        var parentKey = chain.slice(0, -1).join('');
+        currentArrayLength = Array.isArray(val) && val[parentKey] ? val[parentKey].length : 0;
+    }
+
+    var leaf = valuesParsed ? val : parseArrayValue(val, options, currentArrayLength);
 
     for (var i = chain.length - 1; i >= 0; --i) {
         var obj;
@@ -235,6 +259,11 @@ var normalizeParseOptions = function normalizeParseOptions(opts) {
     if (typeof opts.charset !== 'undefined' && opts.charset !== 'utf-8' && opts.charset !== 'iso-8859-1') {
         throw new TypeError('The charset option must be either utf-8, iso-8859-1, or undefined');
     }
+
+    if (typeof opts.throwOnLimitExceeded !== 'undefined' && typeof opts.throwOnLimitExceeded !== 'boolean') {
+        throw new TypeError('`throwOnLimitExceeded` option must be a boolean');
+    }
+
     var charset = typeof opts.charset === 'undefined' ? defaults.charset : opts.charset;
 
     var duplicates = typeof opts.duplicates === 'undefined' ? defaults.duplicates : opts.duplicates;
@@ -266,7 +295,8 @@ var normalizeParseOptions = function normalizeParseOptions(opts) {
         parseArrays: opts.parseArrays !== false,
         plainObjects: typeof opts.plainObjects === 'boolean' ? opts.plainObjects : defaults.plainObjects,
         strictDepth: typeof opts.strictDepth === 'boolean' ? !!opts.strictDepth : defaults.strictDepth,
-        strictNullHandling: typeof opts.strictNullHandling === 'boolean' ? opts.strictNullHandling : defaults.strictNullHandling
+        strictNullHandling: typeof opts.strictNullHandling === 'boolean' ? opts.strictNullHandling : defaults.strictNullHandling,
+        throwOnLimitExceeded: typeof opts.throwOnLimitExceeded === 'boolean' ? opts.throwOnLimitExceeded : false
     };
 };
 
--- a/test/parse.js
+++ b/test/parse.js
@@ -118,7 +118,7 @@ test('parse()', function (t) {
         st.end();
     });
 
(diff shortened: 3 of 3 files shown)


lib/parse.js
```diff
@@ -57,8 +57,19 @@ var parseValues = function parseQueryStringValues(str, options) {
...
+    var parts = cleanStr.split(options.delimiter);
+    
+    // Check if parts exceed the parameter limit and handle it
+    if (parts.length > limit) {
+        if (options.throwOnParameterLimitExceeded) {
+            throw new Error(`Parameter limit exceeded. Only ${limit} parameters allowed.`);
+        }
+        parts = parts.slice(0, limit); // Keep only up to the limit
+    }
+
```
ljharb, line 72:
```suggestion
    var parts = cleanStr.split(
      options.delimiter,
      options.throwOnParameterLimitExceeded ? limit + 1 : limit
    );
    
    // Check if parts exceed the parameter limit and handle it
    if (options.throwOnParameterLimitExceeded && parts.length > limit) {
        throw new Error(`Parameter limit exceeded. Only ${limit} parameters allowed.`);
    }
    
```
this way there's no perf hit on the non-throw path, and it's faster on the throw path (because it only needs to check "limit + 1" (specifically, without the limit i think it would be a DOS vulnerability and require a CVE)

lib/parse.js
```diff
@@ -266,7 +277,8 @@ var normalizeParseOptions = function normalizeParseOptions(opts) {
         parseArrays: opts.parseArrays !== false,
         plainObjects: typeof opts.plainObjects === 'boolean' ? opts.plainObjects : defaults.plainObjects,
         strictDepth: typeof opts.strictDepth === 'boolean' ? !!opts.strictDepth : defaults.strictDepth,
-        strictNullHandling: typeof opts.strictNullHandling === 'boolean' ? opts.strictNullHandling : defaults.strictNullHandling
+        strictNullHandling: typeof opts.strictNullHandling === 'boolean' ? opts.strictNullHandling : defaults.strictNullHandling,
+        throwOnParameterLimitExceeded: typeof opts.throwOnParameterLimitExceeded === 'boolean' ? opts.throwOnParameterLimitExceeded : false
```
ljharb, line 281: we also need to throw if this option is present and is not a boolean

ljharb (changes requested):

lib/parse.js
```diff
@@ -57,8 +57,19 @@ var parseValues = function parseQueryStringValues(str, options) {
...
     var limit = options.parameterLimit === Infinity ? undefined : options.parameterLimit;
-    var parts = cleanStr.split(options.delimiter, limit);
+    var parts = cleanStr.split(
+      options.delimiter,
+      options.throwOnParameterLimitExceeded ? limit + 1 : limit
+    );
+    
+    // Check if parts exceed the parameter limit and handle it
+    if (options.throwOnParameterLimitExceeded && parts.length > limit) {
+        throw new Error(`Parameter limit exceeded. Only ${limit} parameters allowed.`);
```
ljharb, line 70:
```suggestion
        throw new Error('Parameter limit exceeded. Only ' + limit + ' parameters allowed.');
```

README.md
```diff
@@ -135,6 +135,18 @@ var limited = qs.parse('a=b&c=d', { parameterLimit: 1 });
 assert.deepEqual(limited, { a: 'b' });
 ```
 
+If you want an error to be thrown whenever the parameter limit is exceeded, set the `throwOnLimitExceeded` option to `true`. This option will generate a descriptive error if the query string exceeds the parameterLimit.
```
ljharb, line 138:
```suggestion
If you want an error to be thrown whenever the a limit is exceeded (eg, `parameterLimit`, `arrayLimit`), set the `throwOnLimitExceeded` option to `true`. This option will generate a descriptive error if the query string exceeds a configured limit.
```

should we also include `depth` here?

README.md
```diff
@@ -135,6 +135,18 @@ var limited = qs.parse('a=b&c=d', { parameterLimit: 1 });
...
+```javascript
+try {
+    qs.parse('a=1&b=2&c=3&d=4', { parameterLimit: 3, throwOnLimitExceeded: true });
+} catch (err) {
+    assert(err instanceof Error);
+    assert.strictEqual(err.message, 'Parameter limit exceeded. Only 3 parameters allowed.');
+}
+```
+
+When `throwOnLimitExceeded` is set to `false` (default), **qs** will parse up to the specified `parameterLimit` and ignore the rest without throwing an error.
```
ljharb, line 148: account for arrayLimit here also

IamLizu:
@ljharb A question, currently in case `arrayLimit` is exceeded, it is not getting _silently truncated_, right? And we are just throwing the error in case it does?

README.md
IamLizu: Please help me understand about the relation with `depth`.

README.md
IamLizu: A question, currently in case `arrayLimit` is exceeded, it is not getting _silently truncated_, right? And we are just throwing the error in case it does?

For example, If we were to account for `arrayLimit` here, the following test case has to pass,

```js
{
        description: 'throwOnLimitExceeded is set to false, parse upto specified limit and ignore the rest',
        query: 'a[]=1&a[]=2&a[]=3&a[]=4',
        options: { arrayLimit: 5, throwOnLimitExceeded: false },
        expected: { a: ['1', '2', '3'] }
 }
````

But it doesn't, on a limited test run (not a factor), I get,

```console
Test Case: throwOnLimitExceeded is set to false, parse upto specified limit and ignore the rest
Query: a[]=1&a[]=2&a[]=3&a[]=4
Options: { arrayLimit: 5, throwOnLimitExceeded: false }
Parsed Result: { a: [ '1', '2', '3', '4' ] }
Assertion failed: Expected {"a":["1","2","3"]} but got {"a":["1","2","3","4"]}
```

README.md
ljharb: `depth` is also a conceptual limit, that currently when reached, causes a silent behavior change. Throwing would be an alternative.

README.md
ljharb: An arrayLimit of 5 would require 6 items to exceed it; you've only got 4.

IamLizu: Okay, that was a wrong test case, I am not going to prolong that. 

However, I believe array is not truncated to configured limit. In order to test that, I had checked out to main branch and added this test case to the `parse` test cases,

```javascript
t.test('silently truncates array to configured limit', function (sst) {
    const result = qs.parse('a[]=1&a[]=2&a[]=3&a[]=4&a[]=5&a[]=6&a[]=7', { arrayLimit: 5 });

    sst.deepEqual(result, {a: [1, 2, 3, 4, 5] }, 'Should parse and truncate silently');
    sst.end();
});
```

We expect it to pass, but it fails.
```console
# silently truncates array to configured limit
not ok 255 Should parse and truncate silently
  ---
    operator: deepEqual
    expected: |-
      { a: [ 1, 2, 3, 4, 5 ] }
    actual: |-
      { a: [ '1', '2', '3', '4', '5', '6', '7' ] }
```

IamLizu: Do you think it needs to be addressed? I can open a new issue and get started on this.

IamLizu: Do you think it needs to be addressed? I can open a new issue and get started on this.

ljharb: ah, sorry for the confusion. arrayLimit applies with explicit indexes. eg, `qs.parse('a[0]=1&a[10]=2', { arrayLimit: 5 })`.

IamLizu: I have updated the test case to include explicit indices,

```js
t.test('silently truncates array to configured limit', function (sst) {
    const result = qs.parse('a[0]=1&a[1]=2&a[2]=3&a[3]]=4&a[4]=5&a[5]=6&a[6]=7', { arrayLimit: 5 });

    sst.deepEqual(result, {a: [1, 2, 3, 4, 5] }, 'Should parse and truncate silently');
    sst.end();
});
```

I am getting this, 

```console
# silently truncates array to configured limit
not ok 255 Should parse and truncate silently
  ---
    operator: deepEqual
    expected: |-
      { a: [ 1, 2, 3, 4, 5 ] }
    actual: |-
      { a: { 0: '1', 1: '2', 2: '3', 3: '4', 4: '5', 5: '6', 6: '7' } }
```

It seems all seven elements of `a` are getting parsed.

ljharb: right - but the output is an object instead of an array.

the option would cause it to throw when it would fail to produce an array.

IamLizu: Okay

IamLizu: > qs will also limit specifying indices in an array to a maximum index of 20. Any array members with an index of greater than 20 will instead be converted to an object with the index as the key. This is needed to handle cases when someone sent, for example, a[999999999] and it will take significant time to iterate over this huge array.
[qs README](https://github.com/ljharb/qs/blob/main/README.md#parsing-arrays)

Like `parameter` it would not truncate the array silently, rather it creates an object like we have been seeing in above test example. And if the provided array is within limit, then it parses the array in expected output.

So, instead of adding about `throwOnLimitExceeded` effect on `arrayLimit` here, it would be better to add an example in the [parsing-arrays](https://github.com/ljharb/qs/blob/main/README.md#parsing-arrays) section.

README.md
```diff
@@ -286,6 +298,18 @@ var withArrayLimit = qs.parse('a[1]=b', { arrayLimit: 0 });
 assert.deepEqual(withArrayLimit, { a: { '1': 'b' } });
 ```
 
+If you want to throw an error whenever the array limit is exceeded, set the `throwOnLimitExceeded` option to `true`. This option will generate a descriptive error if the query string exceeds a configured limit.
```
IamLizu, line 301: Here's the doc change relavent to `throwOnLimitExceeded` and `arrayLimit`

README.md
IamLizu: There's already a check for `depth`,

```js
if (options.strictDepth === true) {
    throw new RangeError('Input depth exceeded depth option of ' + options.depth + ' and strictDepth is true');
}
```

IamLizu: I am in favor of keeping as is.

lib/parse.js
```diff
@@ -34,11 +34,16 @@ var interpretNumericEntities = function (str) {
...
 
-var parseArrayValue = function (val, options) {
+var parseArrayValue = function (val, options, currentArrayLength) {
     if (val && typeof val === 'string' && options.comma && val.indexOf(',') > -1) {
         return val.split(',');
     }
 
+    // Check if array length exceeds the limit before adding new values
+    if (options.throwOnLimitExceeded && currentArrayLength >= options.arrayLimit) {
+        throw new Error('Array limit exceeded. Only ' + options.arrayLimit + ' elements allowed in an array.');
```
ljharb, line 44:
```suggestion
        throw new Error('Array limit exceeded. Only ' + options.arrayLimit + ' element' + (options.arrayLimit === 1 ? '' : 's') + ' allowed in an array.');
```

lib/parse.js
```diff
@@ -57,8 +62,19 @@ var parseValues = function parseQueryStringValues(str, options) {
...
     var limit = options.parameterLimit === Infinity ? undefined : options.parameterLimit;
-    var parts = cleanStr.split(options.delimiter, limit);
+    var parts = cleanStr.split(
+        options.delimiter,
+        options.throwOnLimitExceeded ? limit + 1 : limit
+    );
+
+    // Check if parts exceed the parameter limit and handle it
+    if (options.throwOnLimitExceeded && parts.length > limit) {
+        throw new Error('Parameter limit exceeded. Only ' + limit + ' parameters allowed.');
```
ljharb, line 75:
```suggestion
        throw new Error('Parameter limit exceeded. Only ' + limit + ' parameter' + (limit === 1 ? '' : 's') + ' allowed.');
```

test/parse.js
```diff
@@ -1035,6 +1035,95 @@ test('parse()', function (t) {
...
+            sst.end();
+        });
+
+        st.test('throws error when array limit exceeded', function (sst) {
+            sst['throws'](
+                function () {
+                    qs.parse('a[]=1&a[]=2&a[]=3&a[]=4', { arrayLimit: 3, throwOnLimitExceeded: true });
+                },
+                new Error('Array limit exceeded. Only 3 elements allowed in an array.'),
+                'Should throw error when array limit is exceeded'
```
ljharb, line 1112:
```suggestion
                'throws error when array limit is exceeded'
```

test/parse.js
```diff
@@ -1035,6 +1035,95 @@ test('parse()', function (t) {
...
+            sst.end();
+        });
+
+        st.test('throws error when throwOnLimitExceeded is present but not boolean for array limit', function (sst) {
+            sst['throws'](
+                function () {
+                    qs.parse('a[]=1&a[]=2&a[]=3&a[]=4', { arrayLimit: 3, throwOnLimitExceeded: 'true' });
+                },
+                new TypeError('`throwOnLimitExceeded` option must be a boolean'),
+                'Should throw error when throwOnLimitExceeded is present and not boolean for array limit'
```
ljharb, line 1101:
```suggestion
                'throws error when throwOnLimitExceeded is present and not boolean for array limit'
```
(same throughout; describe what it does, not what it should do)

test/parse.js
```diff
@@ -1035,6 +1035,95 @@ test('parse()', function (t) {
...
+            sst.end();
+        });
+
+        st.end();
+    });
+
+    t.test('array limit tests', function (st) {
+        st.test('does not throw error when array is within limit', function (sst) {
+            var result = qs.parse('a[]=1&a[]=2&a[]=3', { arrayLimit: 5, throwOnLimitExceeded: true });
+            sst.deepEqual(result, { a: ['1', '2', '3'] }, 'Should parse array without errors');
```
ljharb, line 1091:
```suggestion
            sst.deepEqual(result, { a: ['1', '2', '3'] }, 'parses array without errors');
```

lib/parse.js
```diff
@@ -121,7 +139,13 @@ var parseValues = function parseQueryStringValues(str, options) {
 };
 
 var parseObject = function (chain, val, options, valuesParsed) {
-    var leaf = valuesParsed ? val : parseArrayValue(val, options);
+    var currentArrayLength = 0;
+    if (chain.length > 0 && chain[chain.length - 1] === '[]') {
+        var parentKey = chain.slice(0, -1).join(''); // Get the parent key
```
ljharb, line 144:
```suggestion
        var parentKey = chain.slice(0, -1).join('');
```

ljharb (approved):

README.md
ljharb: I suppose since we have strictDepth already, we don't need the functionality in this option.

PR #530, opened 2025-08-26
Fix parsing of keys containing literal `[]` inside bracket groups

### LINKED ISSUE #493: Incorrect parsing of nested params with closing square bracket ] in property name
Search parameters with closing square bracket in nested property names are parsed incorrectly.

Here demonstrated with `/?search[withbracket[]]=foobar`:

```javascript
> var qs = require("qs");
undefined
> input = { search: { "withbracket[]": "foobar" }}
{ search: { 'withbracket[]': 'foobar' } }
> output = qs.stringify(input, { encode: false })
'search[withbracket[]]=foobar'
> result = qs.parse(output)
{ 'search[withbracket': [ 'foobar' ] }
> assert.deepEqual(input, result)
Uncaught AssertionError [ERR_ASSERTION]: Expected values to be loosely deep-equal:

{
  search: {
    'withbracket[]': 'foobar'
  }
}

should loosely deep-equal

{
  'search[withbracket': [
    'foobar'
  ]
}
    at REPL280:1:8
    at Script.runInThisContext (node:vm:129:12)
    at REPLServer.defaultEval (node:repl:572:29)
    at bound (node:domain:433:15)
    at REPLServer.runBound [as eval] (node:domain:444:12)
    at REPLServer.onLine (node:repl:902:10)
    at REPLServer.emit (node:events:525:35)
    at REPLServer.emit (node:domain:489:12)
    at [_onLine] [as _onLine] (node:internal/readline/interface:425:12)
    at [_line] [as _line] (node:internal/readline/interface:886:18) {
  generatedMessage: true,
  code: 'ERR_ASSERTION',
  actual: [Object],
  expected: [Object],
  operator: 'deepEqual'
}
```

For comparison the result with the native URL object:

```javascript
> url = new URL("http://example.com/?search[withbracket[]]=foobar")
URL {
  href: 'http://example.com/?search[withbracket[]]=foobar',
  origin: 'http://example.com',
  protocol: 'http:',
  username: '',
  password: '',
  host: 'example.com',
  hostname: 'example.com',
  port: '',
  pathname: '/',
  search: '?search[withbracket[]]=foobar',
  searchParams: URLSearchParams { 'search[withbracket[]]' => 'foobar' },
  hash: ''
}
> [...url.searchParams.keys()]
[ 'search[withbracket[]]' ]
```

When parsing a query without braces, the result looks like:

```javascript
> qs.parse("search[okay]=baz")
{ search: { okay: 'baz' } }
```

So it should be a nested object.

### Current Result

```json
{ "search[withbracket": ["foobar"] }
```

### Expected Result

```json
{ "search": { "withbracket[]": ["foobar"] }}
```

techouse:
### Problem

`qs` splits bracket groups with a regex that *forbids* `[]` inside a group:

```js
var child = /(\[[^[][^\]]*])/g; // original: /(\[[^[\]]*])/g
```

This causes keys like `search[withbracket[]]` to split into `"[withbracket]"`, then `"[]"`, leaving the inner `[]` detached from the literal segment. Downstream, it can no longer be treated as a single literal key.

### Repro

```js
// expected:
qs.parse('search[withbracket[]]=foobar');
// => { search: { 'withbracket[]': 'foobar' } }

// actual (today):
// inner [] is split out, so the key can’t be treated as one literal segment
```

### Fix approach

Backport [`splitKeyIntoSegments`](https://github.com/techouse/qs/blob/1952f973a002eae77b08536c6e226603141f5e89/lib/src/extensions/decode.dart#L371) from [qs_dart](https://github.com/techouse/qs) and replace the regex child matcher with a **balanced bracket splitter** that treats `[]` *inside* a bracket group as literal content (only outer `[]` act as array push). This preserves `withbracket[]` as a single segment.

### Notes

- Does not change existing semantics for `[]` at the *outermost* level (array push).
- Honors existing options: `allowDots`, `depth`, `strictDepth`, `allowPrototypes`, etc.

### Test snippet

```js
t.test('parses keys with literal [] inside a bracket group (#493)', function (st) {
    // A bracket pair inside a bracket group should be treated literally as part of the key
    st.deepEqual(
        qs.parse('search[withbracket[]]=foobar'),
        { search: { 'withbracket[]': 'foobar' } },
        'treats inner [] literally when inside a bracket group'
    );

    // Single-level variant
    st.deepEqual(
        qs.parse('a[b[]]=c'),
        { a: { 'b[]': 'c' } },
        'keeps "b[]" as a literal key'
    );

    // Nested with an array push on the outer level
    st.deepEqual(
        qs.parse('list[][x[]]=y'),
        { list: [{ 'x[]': 'y' }] },
        'preserves inner [] while still treating outer [] as array push'
    );

    // Multiple nested bracket pairs: inner [] remains literal as part of the key
    st.deepEqual(
        qs.parse('a[b[c[]]]=d'),
        { a: { 'b[c[]]': 'd' } },
        'treats "b[c[]]" as a literal key inside the bracket group'
    );

    // Depth limits with literal brackets: preserve inner [] while limiting bracket-group parsing
    st.deepEqual(
        qs.parse('a[b[c[]]][d]=e', { depth: 1 }),
        { a: { 'b[c[]]': { '[d]': 'e' } } },
        'respects depth: 1 and preserves literal inner [] in the parsed key'
    );

    // Unterminated inner bracket group is wrapped as a literal remainder segment
    st.deepEqual(
        qs.parse('a[[]b=c'),
        { a: { '[[]b': 'c' } },
        'handles unterminated inner bracket groups without throwing'
    );

    st.end();
});
```

---

Fixes https://github.com/ljharb/qs/issues/493

### DIFF
--- a/lib/parse.js
+++ b/lib/parse.js
@@ -214,9 +214,12 @@ var parseObject = function (chain, val, options, valuesParsed) {
     return leaf;
 };
 
-var splitKeyIntoSegments = function splitKeyIntoSegments(givenKey, options) {
-    var key = options.allowDots ? givenKey.replace(/\.([^.[]+)/g, '[$1]') : givenKey;
+// Split a key like "a[b][c[]]" into ['a', '[b]', '[c[]]'] while preserving
+// qs parse semantics for depth/prototype guards.
+var splitKeyIntoSegments = function splitKeyIntoSegments(originalKey, options) {
+    var key = options.allowDots ? originalKey.replace(/\.([^.[]+)/g, '[$1]') : originalKey;
 
+    // depth <= 0 keeps the whole key as one segment
     if (options.depth <= 0) {
         if (!options.plainObjects && has.call(Object.prototype, key)) {
             if (!options.allowPrototypes) {
@@ -227,14 +230,11 @@ var splitKeyIntoSegments = function splitKeyIntoSegments(givenKey, options) {
         return [key];
     }
 
-    var brackets = /(\[[^[\]]*])/;
-    var child = /(\[[^[\]]*])/g;
+    var segments = [];
 
-    var segment = brackets.exec(key);
-    var parent = segment ? key.slice(0, segment.index) : key;
-
-    var keys = [];
-
+    // parent before the first '[' (may be empty if key starts with '[')
+    var first = key.indexOf('[');
+    var parent = first >= 0 ? key.slice(0, first) : key;
     if (parent) {
         if (!options.plainObjects && has.call(Object.prototype, parent)) {
             if (!options.allowPrototypes) {
@@ -242,32 +242,62 @@ var splitKeyIntoSegments = function splitKeyIntoSegments(givenKey, options) {
             }
         }
 
-        keys[keys.length] = parent;
+        segments[segments.length] = parent;
     }
 
-    var i = 0;
-    while ((segment = child.exec(key)) !== null && i < options.depth) {
-        i += 1;
+    var n = key.length;
+    var open = first;
+    var collected = 0;
 
-        var segmentContent = segment[1].slice(1, -1);
-        if (!options.plainObjects && has.call(Object.prototype, segmentContent)) {
-            if (!options.allowPrototypes) {
-                return;
+    while (open >= 0 && collected < options.depth) {
+        var level = 1;
+        var i = open + 1;
+        var close = -1;
+
+        // balance nested '[' and ']' inside this bracket group using a nesting level counter
+        while (i < n && close < 0) {
+            var cu = key.charCodeAt(i);
+            if (cu === 0x5B) { // '['
+                level += 1;
+            } else if (cu === 0x5D) { // ']'
+                level -= 1;
+                if (level === 0) {
+                    close = i; // found matching close; loop will exit by condition
+                }
             }
+            i += 1;
         }
 
-        keys[keys.length] = segment[1];
+        if (close < 0) {
+            // Unterminated group: wrap the raw remainder in one bracket pair so it stays
+            // a single literal segment (e.g. "[[]b" -> "[[]b]"); we do not infer missing ']'.
+            segments[segments.length] = '[' + key.slice(open) + ']';
+            return segments;
+        }
+
+        var seg = key.slice(open, close + 1);
+        // prototype guard for the content of this group
+        var content = seg.slice(1, -1);
+        if (!options.plainObjects && has.call(Object.prototype, content) && !options.allowPrototypes) {
+            return;
+        }
+
+        segments[segments.length] = seg;
+        collected += 1;
+
+        // find the next '[' after this balanced group
+        open = key.indexOf('[', close + 1);
     }
 
-    if (segment) {
+    if (open >= 0) {
         if (options.strictDepth === true) {
             throw new RangeError('Input depth exceeded depth option of ' + options.depth + ' and strictDepth is true');
         }
 
-        keys[keys.length] = '[' + key.slice(segment.index) + ']';
+        segments[segments.length] = '[' + key.slice(open) + ']';
     }
 
-    return keys;
+    return segments;
 };
 
 var parseKeys = function parseQueryStringKeys(givenKey, val, options, valuesParsed) {
--- a/test/parse.js
+++ b/test/parse.js
@@ -210,6 +210,9 @@ test('parse()', function (t) {
     t.test('uses original key when depth = 0', function (st) {
         st.deepEqual(qs.parse('a[0]=b&a[1]=c', { depth: 0 }), { 'a[0]': 'b', 'a[1]': 'c' });
         st.deepEqual(qs.parse('a[0][0]=b&a[0][1]=c&a[1]=d&e=2', { depth: 0 }), { 'a[0][0]': 'b', 'a[0][1]': 'c', 'a[1]': 'd', e: '2' });
+        st.deepEqual(qs.parse('a.b=c', { depth: 0, allowDots: true }), { 'a[b]': 'c' }, 'normalizes dots before applying depth-0 behavior');
+        st.deepEqual(qs.parse('toString=foo', { depth: 0 }), {}, 'respects prototype guard at depth 0');
+        st.deepEqual(qs.parse('toString=foo', { depth: 0, allowPrototypes: true }), { toString: 'foo' }, 'allows prototypes at depth 0 when enabled');
         st.end();
     });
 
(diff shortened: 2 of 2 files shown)


techouse:
@ljharb looks like some of the tests fail due to the runners being unable to download a particular Node version, i.e.

```
curl: (56) OpenSSL SSL_read: Connection reset by peer, errno 104
download from https://nodejs.org/dist/v5.3.0/node-v5.3.0-linux-x64.tar.xz failed
grep: /home/runner/.nvm/.cache/bin/node-v5.3.0-linux-x64/node-v5.3.0-linux-x64.tar.xz: No such file or directory
```

```
curl: (92) HTTP/2 stream 0 was not closed cleanly: INTERNAL_ERROR (err 2)
download from https://nodejs.org/dist/v13.13.0/node-v13.13.0-linux-x64.tar.xz failed
grep: /home/runner/.nvm/.cache/bin/node-v13.13.0-linux-x64/node-v13.13.0-linux-x64.tar.xz: No such file or directory
```

techouse:
@ljharb, have you had a chance to look at this yet? 😊

ljharb (commented):
This is pretty hard to review. Could you perhaps make a refactoring commit that makes the majority of the changes, and then a fix commit with the test and the thing that makes the test pass?

lib/parse.js
```diff
@@ -187,58 +187,103 @@ var parseObject = function (chain, val, options, valuesParsed) {
...
 };
 
-var parseKeys = function parseQueryStringKeys(givenKey, val, options, valuesParsed) {
-    if (!givenKey) {
-        return;
+// Split a key like "a[b][c[]]" into ['a', '[b]', '[c[]]'] with depth,
+// dot-notation, and strictDepth semantics similar to qs.dart / other ports.
+var splitKeyIntoSegments = function (originalKey, options) {
+    // depth == 0 → no splitting (keeps dots literal, per qs semantics)
+    if (options.depth <= 0) {
```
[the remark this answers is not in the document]
techouse: The pre-refactor code effectively treated `depth <= 0` the same way—by skipping the regex splitting—and never threw. Maintaining that contract avoids surprising callers that rely on the current “no split” semantics for `0` or negative values. Normalisation still coaxes booleans/falsey values elsewhere, so enforcing a new TypeError here would be a regression.

lib/parse.js
```diff
@@ -187,58 +187,103 @@ var parseObject = function (chain, val, options, valuesParsed) {
     return leaf;
 };
 
-var parseKeys = function parseQueryStringKeys(givenKey, val, options, valuesParsed) {
-    if (!givenKey) {
-        return;
+// Split a key like "a[b][c[]]" into ['a', '[b]', '[c[]]'] with depth,
+// dot-notation, and strictDepth semantics similar to qs.dart / other ports.
+var splitKeyIntoSegments = function (originalKey, options) {
+    // depth == 0 → no splitting (keeps dots literal, per qs semantics)
```
ljharb: The original code at `depth <= 0` has a prototype-pollution guard:
```js
if (!options.plainObjects && has.call(Object.prototype, key)) {
    if (!options.allowPrototypes) {
        return;
    }
}
return [key];
```

This replacement drops that guard entirely - when `depth` is 0 and the key matches a property on `Object.prototype` (e.g. `toString`, `constructor`), parsing will no longer be blocked unless `allowPrototypes` is set.

Also, the original returns `[key]` (the whole remaining string as a single segment), but this new code falls through to the regex-based splitting below, which would still split on brackets even at depth 0. That's a behavioral change.

lib/parse.js
```diff
@@ -187,58 +187,103 @@ var parseObject = function (chain, val, options, valuesParsed) {
...
-    var i = 0;
-    while (options.depth > 0 && (segment = child.exec(key)) !== null && i < options.depth) {
-        i += 1;
-        if (!options.plainObjects && has.call(Object.prototype, segment[1].slice(1, -1))) {
-            if (!options.allowPrototypes) {
-                return;
+    // Trailing text after the last balanced group → one final bracket segment (unless it's just '.')
+    if (lastClose >= 0 && lastClose + 1 < n) {
+        var remainder = key.slice(lastClose + 1);
+        if (remainder !== '.') {
```
ljharb: What is the `remainder !== '.'` guard for? When `allowDots` is true, dots are already normalized to bracket notation on line 188 (`givenKey.replace(/\.([^.[]+)/g, '[$1]')`), so `remainder` should never start with `.` at this point.

If this is defending against a specific edge case, could you add a comment or test showing when it triggers?

lib/parse.js
```diff
@@ -187,58 +187,103 @@ var parseObject = function (chain, val, options, valuesParsed) {
...
+            if (options.strictDepth && open >= 0) {
+                throw new RangeError('Input depth exceeded depth option of ' + options.depth + ' and strictDepth is true');
             }
+            segments.push('[' + remainder + ']');
```
ljharb:
```suggestion
            segments[segments.length] = '[' + remainder + ']';
```

ljharb (changes requested):

lib/parse.js
techouse: I restored the original behavior by normalizing `key` first, applying the prototype guard in the `depth <= 0` path, and returning `[key]` (not `[originalKey]`).

lib/parse.js
```diff
@@ -187,58 +187,103 @@ var parseObject = function (chain, val, options, valuesParsed) {
...
+    }
 
-    var i = 0;
-    while (options.depth > 0 && (segment = child.exec(key)) !== null && i < options.depth) {
-        i += 1;
-        if (!options.plainObjects && has.call(Object.prototype, segment[1].slice(1, -1))) {
-            if (!options.allowPrototypes) {
-                return;
+    // Trailing text after the last balanced group → one final bracket segment (unless it's just '.')
+    if (lastClose >= 0 && lastClose + 1 < n) {
```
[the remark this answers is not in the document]
techouse: Agreed.

The `remainder !== '.'` branch was not needed and made behavior less clear.

I removed that special-case and simplified the tail handling to only the depth-overflow remainder path.

lib/parse.js
techouse: Same as https://github.com/ljharb/qs/pull/530#discussion_r2818988369

test/parse.js
```diff
@@ -210,6 +210,9 @@ test('parse()', function (t) {
     t.test('uses original key when depth = 0', function (st) {
         st.deepEqual(qs.parse('a[0]=b&a[1]=c', { depth: 0 }), { 'a[0]': 'b', 'a[1]': 'c' });
         st.deepEqual(qs.parse('a[0][0]=b&a[0][1]=c&a[1]=d&e=2', { depth: 0 }), { 'a[0][0]': 'b', 'a[0][1]': 'c', 'a[1]': 'd', e: '2' });
+        st.deepEqual(qs.parse('a.b=c', { depth: 0, allowDots: true }), { 'a[b]': 'c' }, 'normalizes dots before applying depth-0 behavior');
```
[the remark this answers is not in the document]
techouse: I verified the behavior in the current implementation: with `allowDots: true` and `depth: 0`, the key is normalized first (`a.b` -> `a[b]`) and then returned as a single literal segment in the depth-0 path, so the parsed result is `{ 'a[b]': 'c' }`.

So the test expectation is intentional and matches the actual parser behavior.

test/parse.js
```diff
@@ -210,6 +210,9 @@ test('parse()', function (t) {
...
+        st.deepEqual(qs.parse('toString=foo', { depth: 0 }), {}, 'respects prototype guard at depth 0');
+        st.deepEqual(qs.parse('toString=foo', { depth: 0, allowPrototypes: true }), { toString: 'foo' }, 'allows prototypes at depth 0 when enabled');
```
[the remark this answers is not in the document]
techouse: These assertions are intentional and in scope for this PR.

The refactor changed `splitKeyIntoSegments`, including the `depth <= 0` path where prototype-guard behavior is applied. These tests are regression coverage to ensure that behavior remains correct after the bracket-parsing changes, so keeping them alongside other depth=0 parse tests is appropriate.

ljharb (approved):

Ari4ka (approved):

MortimerSC (commented):
Fixes #558

Issue #537, opened 2026-01-02
Query string array parsing behavior changed in version 6.14.1


francescorivola:
Follow test is passing in version 6.14.0, but it fails in 6.14.1.

```js
import { deepEqual } from "node:assert/strict";
import { describe, it } from "node:test";
import qs from "qs";

describe("qs module", function () {
  it("should parse query strings as an array", function () {
    const values = Array.from({ length: 21 }, (_, index) => "value" + index);
    const queryString = values.map((id) => `id=${id}`).join("&");
    const parsed = qs.parse(queryString);
    deepEqual(parsed.id, values);
  });
});

```
Likely related to https://github.com/ljharb/qs/commit/3086902ecf7f088d0d1803887643ac6c03d415b9

After upgrade due to an npm audit report we have observed test failures. Our express application is allowing up to 50 ids in an endpoint query string param. I am not sure if was intended but TBH I think it could be dangerous introducing this behavior change as part of a patch version.

Please, let me know if you need any other information, glad to help.

Thank you very much.

kroleg:
But `qs` is usually a nested dependency so there is no way to override `arrayLimit` because qs.parse is called somewhere in third party code. Also there is no way to change defaults for `qs` and even if there was one different libs can depend on different versions of `qs` so there is no easy way to set defaults for all.

I see 2 possible solutions:
1. Raise default from `20` to `100` or even bigger (original [commit](https://github.com/ljharb/qs/commit/4667c41c586ba2bca3c2fdc8c7cc362587ddeeae) doesn't have any justification for value). Not a solid solution because there might still be people who need bigger value.
2. (solid) Add a way to override it globally via env var (making sure it doesn't fail in non-node.js runtimes)
```js

  function getEnvVar (name) {
      try {
          // Node.js / Bun
          if (typeof process !== 'undefined' && process.env) {
              return process.env[name];
          }
          // Deno
          if (typeof Deno !== 'undefined' && Deno.env && typeof Deno.env.get === 'function') {
              return Deno.env.get(name);
          }
      } catch (e) {
          // Deno throws without --allow-env permission, ignore
          // can probably log a warning
      }
      return undefined;
  }

  function parseEnvInt (val, fallback) {
      if (val === undefined || val === null || val === '') {
          return fallback;
      }
      var parsed = parseInt(val, 10);
      return isNaN(parsed) ? fallback : parsed;
  };

  var defaults = {
[...]

PR #538, opened 2026-01-05
Correct changelog entry for arrayLimit fix

kroleg:
- should mention `arrayLimit` because it's what devs use and `arrayLength` is internal variable
- should warn that it is a breaking change (example of people reporting behavior change: https://github.com/ljharb/qs/issues/537)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,5 @@
 ## **6.14.1**
-- [Fix] ensure arrayLength applies to `[]` notation as well
+- [Fix] (breaking change) ensure arrayLimit applies to `[]` notation as well
 - [Fix] `parse`: when a custom decoder returns `null` for a key, ignore that key
 - [Refactor] `parse`: extract key segment splitting helper
 - [meta] add threat model


CHANGELOG.md
```diff
@@ -1,5 +1,5 @@
 ## **6.14.1**
-- [Fix] ensure arrayLength applies to `[]` notation as well
+- [Fix] (breaking change) ensure arrayLimit applies to `[]` notation as well
```
ljharb: it is absolutely not a breaking change; bugfixes can sometimes break people if they're depending on broken behavior.

ljharb (changes requested):

CHANGELOG.md
kroleg: Imo it doesn't matter if behaviour was broken or not. What matters is that this is a significant behaviour change.

Idk but feels like you're being defensive or smth. It's not like i am blaming you for broken apps (it wouldn't make any sense) or asking to release new major version because of it. It's me trying to warn other devs, so people updating to 6.14.1 will be aware that it can break their apps.

ljharb: If I agreed it was a breaking change, then I would be compelled to revert it, and publish a new major, and tons of people on v6 would be left vulnerable.

I don't agree it's a breaking change, which is why it's fine to release it in a non-major. EVERY change _could_ break someone's app; that's what tests and lockfiles are for.

kroleg: Got you. What wording would allow to warn people but not force you to publish new major? Would you be ok with something like: 
```
- [Fix] ensure arrayLimit applies to `[]` notation as well. 
  WARN: if you handle more than `arrayLimit` (20 by default) items in array query param (i.e. `param[]=a&...param[]=x`) it will be converted to object instead of array
```

ljharb: I don't think any warning is required, personally.

ljharb:
This was fixed in 2a35775.

PR #542, opened 2026-01-20
add allowDotsForNumbers option to convert from brackets of indices to dots

### LINKED ISSUE #516: stringify, request supporting 'dot' as option for arrayFormat
parse can successfully convert 'a.0.name=test&a.0.number=5&a.1.name=test2&a.1.number=6' to: 

{ a: [ { name: 'test', number: '5'}, { name: 'test2', number: '6' }] }

so it would be great if there were a stringify option that could convert the object back into the original dot notation.

antareepsarkar:
resolves #516
 
`allowDots` does not convert bracket notation for array indices to dots. 
This PR adds an option `allowDotsForIndices` which does that.

### DIFF
--- a/lib/parse.js
+++ b/lib/parse.js
@@ -7,6 +7,7 @@ var isArray = Array.isArray;
 
 var defaults = {
     allowDots: false,
+    allowDotsForIndices: false,
     allowEmptyArrays: false,
     allowPrototypes: false,
     allowSparse: false,
@@ -215,7 +216,8 @@ var parseObject = function (chain, val, options, valuesParsed) {
 };
 
 var splitKeyIntoSegments = function splitKeyIntoSegments(givenKey, options) {
-    var key = options.allowDots ? givenKey.replace(/\.([^.[]+)/g, '[$1]') : givenKey;
+    var key = options.allowDots
+        ? options.allowDotsForIndices ? givenKey.replace(/\.([^.[]+)/g, '[$1]') : givenKey.replace(/\.(?!\d)([^.[\]]+)/g, '[$1]') : givenKey;
 
     if (options.depth <= 0) {
         if (!options.plainObjects && has.call(Object.prototype, key)) {
@@ -317,10 +319,17 @@ var normalizeParseOptions = function normalizeParseOptions(opts) {
         throw new TypeError('The duplicates option must be either combine, first, or last');
     }
 
-    var allowDots = typeof opts.allowDots === 'undefined' ? opts.decodeDotInKeys === true ? true : defaults.allowDots : !!opts.allowDots;
+    if ('allowDots' in opts && opts.allowDots === false && opts.allowDotsForIndices) {
+        throw new TypeError('allowDots cannot be false and allowDotsForIndices true at the same time');
+    }
+
+    var allowDots = typeof opts.allowDots === 'undefined' ? opts.decodeDotInKeys === true || !!opts.allowDotsForIndices
+        ? true : defaults.allowDots : !!opts.allowDots;
+    var allowDotsForIndices = typeof opts.allowDotsForIndices === 'undefined' ? allowDots : !!opts.allowDotsForIndices;
 
     return {
         allowDots: allowDots,
+        allowDotsForIndices: allowDotsForIndices,
         allowEmptyArrays: typeof opts.allowEmptyArrays === 'boolean' ? !!opts.allowEmptyArrays : defaults.allowEmptyArrays,
         allowPrototypes: typeof opts.allowPrototypes === 'boolean' ? opts.allowPrototypes : defaults.allowPrototypes,
         allowSparse: typeof opts.allowSparse === 'boolean' ? opts.allowSparse : defaults.allowSparse,
--- a/lib/parse.js
+++ b/lib/parse.js
@@ -7,7 +7,7 @@ var isArray = Array.isArray;
 
 var defaults = {
     allowDots: false,
-    allowDotsForIndices: false,
+    allowDotsForNumbers: false,
     allowEmptyArrays: false,
     allowPrototypes: false,
     allowSparse: false,
@@ -216,8 +216,8 @@ var parseObject = function (chain, val, options, valuesParsed) {
 };
 
 var splitKeyIntoSegments = function splitKeyIntoSegments(givenKey, options) {
-    var key = options.allowDots
-        ? options.allowDotsForIndices ? givenKey.replace(/\.([^.[]+)/g, '[$1]') : givenKey.replace(/\.(?!\d)([^.[\]]+)/g, '[$1]') : givenKey;
+    var key = options.allowDots ? options.allowDotsForNumbers
+        ? givenKey.replace(/\.([^.[]+)/g, '[$1]') : givenKey.replace(/\.(?!\d+(?:\.|\[|\]|$))([^.[]+)/g, '[$1]') : givenKey;
 
     if (options.depth <= 0) {
         if (!options.plainObjects && has.call(Object.prototype, key)) {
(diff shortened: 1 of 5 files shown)


Pratikshit26:
pretest nodes please specify

Pratikshit26:
Automatic rebase ??

ljharb (changes requested):
what about for parse?

lib/parse.js
```diff
@@ -203,7 +204,8 @@ var parseObject = function (chain, val, options, valuesParsed) {
 };
 
 var splitKeyIntoSegments = function splitKeyIntoSegments(givenKey, options) {
-    var key = options.allowDots ? givenKey.replace(/\.([^.[]+)/g, '[$1]') : givenKey;
+    var key = options.allowDots
+        ? options.allowDotsForIndices ? givenKey.replace(/\.([^.[]+)/g, '[$1]') : givenKey.replace(/\.(?!\d)([^.[\]]+)/g, '[$1]') : givenKey;
```
ljharb: this regex differs from the original in the capture group: it adds ] to the exclusion set ([^.[\]] vs [^.[). This means keys containing ] that were previously captured now won't be. This would be a breaking change.

additionally, the negative lookahead (?!\d) is too broad. It prevents ANY dot-before-digit from being treated as property access, not just dots before valid array indices. For example, a.1foo.b=c - 1foo is not a valid array index, but the regex would still skip .1foo, producing `{ 'a.1foo': { b: 'c' } }` instead of the current `{ a: { '1foo': { b: 'c' } } }`, which is another breaking change.

ljharb (changes requested):
Note that the issue suggests "allowDotsForNumbers" - "for indices" and "for numbers" are distinctly different and imply different semantics.

Additionally, the intention was for `allowDots` to be *required to be explicitly passed as true* when the new option is passed.

Also, if this new option is omitted, behavior MUST not change, and in this PR, it does.

If allowDotsForIndices: true and encodeDotInKeys: true, the dots used for indices would get encoded as %2E, defeating the purpose. There's no guard for this.

lib/parse.js
```diff
@@ -215,7 +216,8 @@ var parseObject = function (chain, val, options, valuesParsed) {
 };
 
 var splitKeyIntoSegments = function splitKeyIntoSegments(givenKey, options) {
-    var key = options.allowDots ? givenKey.replace(/\.([^.[]+)/g, '[$1]') : givenKey;
+    var key = options.allowDots ? options.allowDotsForNumbers
+        ? givenKey.replace(/\.([^.[]+)/g, '[$1]') : givenKey.replace(/\.(?!\d+(?:\.|\[|\]|$))([^.[]+)/g, '[$1]') : givenKey;
```
ljharb, line 220: Triple-nested ternary across two lines is hard to follow, and the line break is in the middle of a ternary branch. Two named regex constants and a flat `if`/`else if`/`else` (or at minimum properly grouped parens) would read much more clearly.

lib/parse.js
```diff
@@ -317,10 +319,17 @@ var normalizeParseOptions = function normalizeParseOptions(opts) {
         throw new TypeError('The duplicates option must be either combine, first, or last');
     }
 
-    var allowDots = typeof opts.allowDots === 'undefined' ? opts.decodeDotInKeys === true ? true : defaults.allowDots : !!opts.allowDots;
+    if ('allowDots' in opts && opts.allowDots === false && opts.allowDotsForNumbers) {
+        throw new TypeError('allowDots cannot be false and allowDotsForNumbers true at the same time');
+    }
+
+    var allowDots = typeof opts.allowDots === 'undefined' ? opts.decodeDotInKeys === true || !!opts.allowDotsForNumbers
+        ? true : defaults.allowDots : !!opts.allowDots;
```
ljharb, line 327: Per the previous review, the intention was to require `allowDots` to be explicitly `true` when `allowDotsForNumbers` is passed. This still auto-enables `allowDots` instead. Suggest a `TypeError` when `opts.allowDotsForNumbers === true` and `opts.allowDots !== true`, mirroring the new throw above.

lib/stringify.js
```diff
@@ -10,6 +10,9 @@ var arrayPrefixGenerators = {
         return prefix + '[]';
     },
     comma: 'comma',
+    dots: function dots(prefix, key) {
```
ljharb, line 13: Adding `dots` to `arrayPrefixGenerators` makes `arrayFormat: 'dots'` a public option. If that's intentional, it should be documented in the README alongside `indices`/`brackets`/`repeat`/`comma`. If it isn't intentional, gate it behind `allowDotsForNumbers` instead of putting it in this map.

lib/stringify.js
```diff
@@ -220,6 +226,10 @@ var normalizeStringifyOptions = function normalizeStringifyOptions(opts) {
         throw new TypeError('Encoder has to be a function.');
     }
 
+    if ('allowDots' in opts && opts.allowDots === false && opts.allowDotsForNumbers) {
```
ljharb, line 229: Missing the second guard from the previous review: `encodeDotInKeys: true` plus `allowDotsForNumbers: true` would percent-encode the index dots to `%2E`, defeating the feature. Should throw here as well.

lib/stringify.js
```diff
@@ -244,6 +254,8 @@ var normalizeStringifyOptions = function normalizeStringifyOptions(opts) {
         arrayFormat = opts.arrayFormat;
     } else if ('indices' in opts) {
         arrayFormat = opts.indices ? 'indices' : 'repeat';
+    } else if (opts.allowDotsForNumbers) {
```
ljharb, line 257: Implicitly switching `arrayFormat` to `'dots'` based on `allowDotsForNumbers` may surprise users; better to require an explicit `arrayFormat: 'dots'` and have `allowDotsForNumbers` only affect dot-vs-bracket handling for numeric segments, the same way `allowDots` does for non-numeric ones.

lib/stringify.js
```diff
@@ -253,10 +265,12 @@ var normalizeStringifyOptions = function normalizeStringifyOptions(opts) {
     }
 
     var allowDots = typeof opts.allowDots === 'undefined' ? opts.encodeDotInKeys === true ? true : defaults.allowDots : !!opts.allowDots;
+    var allowDotsForNumbers = !!opts.allowDotsForNumbers;
```
ljharb, line 268: Asymmetric with parse, where `allowDotsForNumbers` defaults to `allowDots`. Result: `qs.stringify({a:[1,2]}, { allowDots: true })` does not produce `a.0=1&a.1=2`, even though `qs.parse('a.0=1', { allowDots: true })` treats `0` as an array index. Round-trip is broken unless the caller sets both. Suggest matching parse: default this to `allowDots`.

test/parse.js
```diff
@@ -63,6 +63,12 @@ test('parse()', function (t) {
         st.end();
     });
 
+    t.test('allows enabling dots for numbers', function (st) {
+        st.deepEqual(qs.parse('a.0.b=d&a.1.c=e', { allowDotsForNumbers: false, allowDots: true }), { 'a.0': { b: 'd' }, 'a.1': { c: 'e' } }, 'with allowDotsForNumbers false and allowDots true');
+        st.deepEqual(qs.parse('a.b.0=c&a.c.0=d', { allowDotsForNumbers: true }), { a: { b: ['c'], c: ['d'] } }, 'with only allowDotsForIndices true');
```
ljharb, line 68: Stale label: option was renamed `allowDotsForNumbers`, but this assertion message still says `allowDotsForIndices`. The PR title also still uses "indices".

ljharb (changes requested):
Thanks for the iteration. A few items from the earlier review still aren't addressed; see inline comments, but the big ones are that `allowDots: true` should be required (not auto-enabled) when `allowDotsForNumbers` is set, the `encodeDotInKeys` + `allowDotsForNumbers` combo needs a guard on stringify, and the naming is inconsistent (code uses `allowDotsForNumbers`, but the PR title and one test label still say `allowDotsForIndices`). Also needs a rebase against current `main`, plus README/changelog updates.

PR #550, opened 2026-03-13
[readme]: Minor change of grammar

JoBeGaming:
Use imperative in README, similar to usage of such explained in the [Commit Message Guidelines](https://gist.github.com/ljharb/772b0334387a4bee89af24183114b3c7).


ljharb (approved):
this isn't about imperative or not, to be clear, this is just incorrect grammar. thanks for the fix

PR #551, opened 2026-04-13
[Fix] `stringify`: skip null/undefined filter-array entries instead of crashing in `encoder`

gregkh:
When `filter` is an array, its elements become `objKeys` directly and the top-level loop passes each entry as the recursive `prefix` argument without coercion.  A `null` or `undefined` entry reached `utils.encode` as the `str` argument and `str.length` threw `TypeError`.

The crash requires two conditions to align: the filter array must contain `null`/`undefined`, and the object being stringified must have a literal `'null'`/`'undefined'` property (otherwise `obj[key]` is `undefined` and the typeof-undefined early-return at stringify.js:137 fires before the encoder is called).  Both are reachable from an attacker controlled input in applications that stringify user-submitted JSON with user-selected fields.

Fix this up by skipping `null`/`undefined` filter entries at the top-level loop.  This matches `JSON.stringify` replacer-array semantics, which silently ignore non-string/non-number entries.  `Object.keys` never yields `null`/`undefined`, so the guard only affects user-supplied filter arrays.  The inner recursive loop already coerces `key` via `String()` when building `keyPrefix`, so it never passes a raw `null` to the encoder.

### DIFF
--- a/lib/stringify.js
+++ b/lib/stringify.js
@@ -312,6 +312,17 @@ module.exports = function (object, opts) {
     var sideChannel = getSideChannel();
     for (var i = 0; i < objKeys.length; ++i) {
         var key = objKeys[i];
+
+        // Skip null/undefined entries when objKeys came from a filter array.
+        // Without this guard, the raw key is passed as the prefix to the
+        // recursive stringify call below, and encoder(null).length throws.
+        // Object.keys never yields null/undefined, so this only affects
+        // user-supplied filter arrays — matching JSON.stringify replacer
+        // semantics, which silently ignore non-string/number entries.
+        if (typeof key === 'undefined' || key === null) {
+            continue;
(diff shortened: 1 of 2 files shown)


ljharb (changes requested):
This would need regression tests.

gregkh:
Regression test added, thanks for reminding me!

ljharb (approved):
Although I'd prefer you didn't use LLMs for open source contributions, this seems fine, thanks.

gregkh:
> Although I'd prefer you didn't use LLMs for open source contributions, this seems fine, thanks.

Sorry about that, a LLM found the problem, and helped with a portion of this.  If there is a better way to mark this as such, I'll be glad to do so.

Thanks for taking the change, much appreciated.

PR #552, opened 2026-04-16
[Perf]: ~1.9x faster parse via fast-paths and manual scanning

homanp:
## Summary

~1.9x faster `qs.parse` via fast-paths for common cases and manual scanning to replace split/regex in hot loops. No behavior changes, no new deps. All 893 tests pass.

## Benchmarks

Weighted throughput across 5 representative query string shapes, 3 fresh node processes (Node 23, macOS):

| | baseline | this PR | speedup |
|---|---|---|---|
| flat (`a=1&b=2&c=3&d=4&e=5`) | 667k | 1,540k | 2.3x |
| nested (`user[name]=alice&...`) | 355k | 496k | 1.4x |
| arrays (`ids[]=1&ids[]=2&...`) | 353k | 528k | 1.5x |
| encoded (`q=hello%20world&...`) | 848k | 1,365k | 1.6x |
| search (multi-field form) | 322k | 607k | 1.9x |
| **weighted** | **506k** | **947k** | **1.9x** |

Bench script in `bench/workload.js` for reproduction.

## What changed

1. **Flat-key fast-path**: when a key has no `[`, skip bracket-parsing entirely.
2. **Decoder skip**: `indexOf('%')` + `indexOf('+')` up front — if neither present, return input unchanged.
3. **Manual delimiter scan**: for single-char delimiters (default `&`), in-place scan instead of `split()`.
4. **Skip compact for flat-only**: `compact()` only needed when bracket parsing created sparse arrays.
5. **Manual bracket scanner**: replaced bracket-matching regexes with `indexOf` in `splitKeyIntoSegments`.

## Correctness

All 893 tape tests pass. Equivalence also verified on 25 additional adversarial inputs (empty strings, bare keys, duplicates, deep nesting, percent-encoded specials, `+` as space, mixed bracket styles).

Happy to split into separate commits per optimization if that helps review.

### DIFF
--- a/lib/parse.js
+++ b/lib/parse.js
@@ -61,14 +61,33 @@ var charsetSentinel = 'utf8=%E2%9C%93'; // encodeURIComponent('✓')
 var parseValues = function parseQueryStringValues(str, options) {
     var obj = { __proto__: null };
 
-    var cleanStr = options.ignoreQueryPrefix ? str.replace(/^\?/, '') : str;
-    cleanStr = cleanStr.replace(/%5B/gi, '[').replace(/%5D/gi, ']');
+    var cleanStr = options.ignoreQueryPrefix && str.charCodeAt(0) === 0x3F ? str.slice(1) : str;
+    if (cleanStr.indexOf('%5') !== -1) {
+        cleanStr = cleanStr.replace(/%5B/gi, '[').replace(/%5D/gi, ']');
+    }
 
     var limit = options.parameterLimit === Infinity ? void undefined : options.parameterLimit;
-    var parts = cleanStr.split(
-        options.delimiter,
-        options.throwOnLimitExceeded && typeof limit !== 'undefined' ? limit + 1 : limit
-    );
+    var parts;
+    var delimiter = options.delimiter;
+    if (typeof delimiter === 'string' && delimiter.length === 1) {
+        parts = [];
+        var splitLimit = typeof limit === 'undefined' ? Infinity : options.throwOnLimitExceeded ? limit + 1 : limit;
+        var start = 0;
+        var idx = cleanStr.indexOf(delimiter);
+        while (idx !== -1 && parts.length < splitLimit) {
+            parts[parts.length] = cleanStr.slice(start, idx);
+            start = idx + 1;
+            idx = cleanStr.indexOf(delimiter, start);
+        }
+        if (parts.length < splitLimit) {
+            parts[parts.length] = cleanStr.slice(start);
+        }
+    } else {
+        parts = cleanStr.split(
+            delimiter,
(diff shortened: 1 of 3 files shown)


lib/parse.js
```diff
@@ -61,14 +61,33 @@ var charsetSentinel = 'utf8=%E2%9C%93'; // encodeURIComponent('✓')
 var parseValues = function parseQueryStringValues(str, options) {
     var obj = { __proto__: null };
 
-    var cleanStr = options.ignoreQueryPrefix ? str.replace(/^\?/, '') : str;
-    cleanStr = cleanStr.replace(/%5B/gi, '[').replace(/%5D/gi, ']');
+    var cleanStr = options.ignoreQueryPrefix && str.charCodeAt(0) === 0x3F ? str.slice(1) : str;
+    if (cleanStr.indexOf('%5') !== -1) {
+        cleanStr = cleanStr.replace(/%5B/gi, '[').replace(/%5D/gi, ']');
+    }
```
ljharb: i'm surprised "indexOf + 2 replaces" being slower doesn't outweigh the benefit of the other branch being "indexOf instead of 2 replaced". it might be a good idea to do each change in a separate commit, and put the benchmark results in each commit message.

ljharb (changes requested):
This will need a rebase; it conflicts with some bugfixes i merged.

lib/parse.js
homanp: Ageee, Will make it clearer. Ty for feedback.

homanp:
@ljharb Rebased and split into separate commits with benchmarks in each.

After testing each optimization in isolation I ended up dropping two of them: 

- The indexOf('%5') guard before the %5B/%5D replace, you were right to flag it, only ~1% gain end-to-end. Just not worth it imo. 
- The manual indexOf delimiter splitting loop, ~2.5% gain for 17 extra lines. Same here not worth it.

lib/parse.js
homanp: Done.

techouse:
FWIW, I maintain a Dart port that stays very close to Node `qs` semantics, and the decode optimization work in techouse/qs#51 ended up in a similar direction: the biggest wins came from guarded fast paths for flat query strings, avoiding key splitting/merge/compact when no structured syntax is present, and skipping decode work for tokens that contain neither `%` nor `+`.

I would not treat the Dart benchmark numbers as portable to V8, but the regression set was useful. The fast path had to prove behavior for mixed flat/structured keys (`a=1&a[b]=2`, `a[b]=2&a=1`), duplicates + comma, charset sentinel, `allowDots`/encoded-dot cases, prototype guards, array/parameter limits, and the recent balanced-bracket parsing cases.

PR #554, opened 2026-04-24
[Fix] `stringify`: apply `formatter` to encoded key under `strictNullHandling`

deepview-autofix:
Previously the null branch returned the raw `encoder(prefix, ...)` output (or raw `prefix`) without running it through `formatter`. With `format: 'RFC1738'` this caused keys containing spaces to serialize as `%20` in strict-null mode while the normal primitive path produced `+`, yielding inconsistent output for the same key.

### DIFF
--- a/lib/stringify.js
+++ b/lib/stringify.js
@@ -118,7 +118,7 @@ var stringify = function stringify(
 
     if (obj === null) {
         if (strictNullHandling) {
-            return encoder && !encodeValuesOnly ? encoder(prefix, defaults.encoder, charset, 'key', format) : prefix;
+            return formatter(encoder && !encodeValuesOnly ? encoder(prefix, defaults.encoder, charset, 'key', format) : prefix);
         }
 
         obj = '';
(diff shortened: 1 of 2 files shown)


ChALkeR (commented):
Before the fix:
```
> require('qs').stringify({'a a':null,'a b':1},{format:'RFC1738',strictNullHandling:true})
'a%20a&a+b=1'
```

Had both `%20` and `+`.

lib/stringify.js
```diff
@@ -118,7 +118,7 @@ var stringify = function stringify(
 
     if (obj === null) {
         if (strictNullHandling) {
-            return encoder && !encodeValuesOnly ? encoder(prefix, defaults.encoder, charset, 'key', format) : prefix;
+            return encoder && !encodeValuesOnly ? formatter(encoder(prefix, defaults.encoder, charset, 'key', format)) : formatter(String(prefix));
```
ChALkeR:
```suggestion
            return formatter(encoder && !encodeValuesOnly ? encoder(prefix, defaults.encoder, charset, 'key', format) : prefix);
```

ljharb (approved):
added an extra assertion with the same input, with the opposite strictNullHandling option

PR #555, opened 2026-04-24
[Fix] `stringify`: use configured `delimiter` after `charsetSentinel`

deepview-autofix:
The `utf8=...` charset sentinel was followed by a hardcoded `&`, ignoring `options.delimiter`. With any non-`&` delimiter the output mixed delimiters (e.g. `utf8=%E2%9C%93&a=1;b=2`), breaking self round-tripping. Emit the configured delimiter instead, and add a regression test.


ljharb (approved):

AmedeoPelliccia:
And thnks

PR #556, opened 2026-04-29
fix: consistent array overflow handling regardless of arrayLimit value

### LINKED ISSUE #546: When `arrayLimit` is exceeded and mixed array notation is used, result depends on the limit value
This is something that I found in https://github.com/ljharb/qs/issues/540#issuecomment-3764695118, but was not fixed by #545 (fbc5206c25b4d1851cea683f02c10756c521d15a). Given a string with mixed array notation `a[]=b&a[1]=c&a=d` and `arrayLimit` values `-1`, `0` and `1` (the limit is always exceeded), `qs.parse()` returns 3 different results:

| input | `arrayLimit` | result |
| - | - | - |
| `a[]=b&a[1]=c&a=d` | -1 | `{ a: { '0': 'b', '1': 'd' } }` |
| `a[]=b&a[1]=c&a=d` | 0 | `{ a: { '0': 'b', '1': 'c', d: true } }` |
| `a[]=b&a[1]=c&a=d` | 1 | `{ a: { '0': 'b', '1': 'c', '2': 'd' } }` |

This has something to do with the order in which everything is combined and when exactly is the overflow detected. When trying to debug it previously, I found out that there is a case, when object is [not marked](https://github.com/ljharb/qs/blob/f6a7abff1f13d644db9b05fe4f2c98ada6bf8482/lib/parse.js#L201-L203) as an overflow object when the overflow [is detected](https://github.com/ljharb/qs/blob/f6a7abff1f13d644db9b05fe4f2c98ada6bf8482/lib/parse.js#L197):

https://github.com/ljharb/qs/blob/f6a7abff1f13d644db9b05fe4f2c98ada6bf8482/lib/parse.js#L197-L203

This is also possibly related to https://github.com/ljharb/qs/issues/529#issuecomment-3342023154.

abhu85:
## Summary
Fix inconsistent array overflow handling when `arrayLimit` is exceeded with mixed array notation.

## Problem
Parsing `a[]=b&a[1]=c&a=d` produces different results depending on the `arrayLimit` value (-1, 0, 1), even though the limit is exceeded in all cases. For example, `a[0]=c&a=b` with `arrayLimit: 1` produced `["c","b"]` (array) while `arrayLimit: 0` and `-1` both produced `{0:"c", 1:"b"}` (object).

The overflow detection in `lib/utils.js` `merge` function used `>` (strict greater-than) instead of `>=` when checking whether appending to an array would exceed the limit, allowing arrays at exactly the boundary to grow past the limit.

## Solution
Changed the overflow boundary check in two places in `utils.js` `merge` from `>` to `>=`:
1. When merging a non-object source into an array target (line 78)
2. When merging a non-object target with an array/object source (line 117)

This ensures that when an array is already at its maximum allowed length (`arrayLimit`), adding more elements triggers overflow conversion to an object, consistent with how lower `arrayLimit` values behave.

## Test Plan
- [x] `a[]=b&a[1]=c&a=d` with arrayLimit -1, 0, 1 all produce `{0:'b', 1:'c', 2:'d'}`
- [x] `a[0]=c&a=b` with arrayLimit -1, 0, 1 all produce `{0:'c', 1:'b'}`
- [x] `a=b&a[0]=c` with arrayLimit -1, 0, 1 all produce `{0:'b', 1:'c'}`
- [x] Normal array parsing within limits still works
- [x] Existing test suite passes (929/929) with 100% coverage

Fixes #546

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -75,7 +75,7 @@ var merge = function merge(target, source, options) {
     if (typeof source !== 'object' && typeof source !== 'function') {
         if (isArray(target)) {
             var nextIndex = target.length;
-            if (options && typeof options.arrayLimit === 'number' && nextIndex > options.arrayLimit) {
+            if (options && typeof options.arrayLimit === 'number' && nextIndex >= options.arrayLimit) {
                 return markOverflow(arrayToObject(target.concat(source), options), nextIndex);
             }
             target[nextIndex] = source;
@@ -114,7 +114,7 @@ var merge = function merge(target, source, options) {
             return markOverflow(result, getMaxIndex(source) + 1);
         }
         var combined = [target].concat(source);
-        if (options && typeof options.arrayLimit === 'number' && combined.length > options.arrayLimit) {
+        if (options && typeof options.arrayLimit === 'number' && combined.length >= options.arrayLimit) {
             return markOverflow(arrayToObject(combined, options), combined.length - 1);
         }
         return combined;
--- a/test/parse.js
+++ b/test/parse.js
@@ -555,6 +555,73 @@ test('parse()', function (t) {
         st.end();
     });
 
+    t.test('consistent overflow handling regardless of arrayLimit value', function (st) {
+        st.deepEqual(
+            qs.parse('a[]=b&a[1]=c&a=d', { arrayLimit: -1 }),
+            { a: { 0: 'b', 1: 'c', 2: 'd' } },
+            'mixed array notation with arrayLimit -1 produces object'
+        );
+        st.deepEqual(
+            qs.parse('a[]=b&a[1]=c&a=d', { arrayLimit: 0 }),
+            { a: { 0: 'b', 1: 'c', 2: 'd' } },
+            'mixed array notation with arrayLimit 0 produces object'
+        );
+        st.deepEqual(
+            qs.parse('a[]=b&a[1]=c&a=d', { arrayLimit: 1 }),
+            { a: { 0: 'b', 1: 'c', 2: 'd' } },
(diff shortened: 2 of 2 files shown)


ljharb:
I added test cases with arrayLimit 2. 4 tests now fail without the fix (up from 1) and 1 test fails with the fix.

That 1 test could easily be made to pass in both (ie, to make 3 failures vs 0) but I'm concerned about the scope of this change. If it occurs with any arrayLimit value, it could impact a lot of users who are unintentionally relying on the current behavior.

abhu85:
You're right to be concerned — I dug into this more carefully and I think my `>=` change is wrong.

The original inconsistency from #546 (different results for `arrayLimit` -1, 0, and 1 with `a[]=b&a[1]=c&a=d`) is already fixed on current `main`, likely by your Feb 2026 commits (`cfc108f`, `6addf8c`). I verified:

```
// On current main WITHOUT my >= change — all consistent:
a[]=b&a[1]=c&a=d, limit -1 → {0:'b', 1:'c', 2:'d'}
a[]=b&a[1]=c&a=d, limit  0 → {0:'b', 1:'c', 2:'d'}
a[]=b&a[1]=c&a=d, limit  1 → {0:'b', 1:'c', 2:'d'}
```

The `>=` change effectively shifts the arrayLimit boundary down by 1 for *all* values, which means `arrayLimit: N` starts rejecting arrays that should be valid (e.g., `a[0]=c&a=b` with `arrayLimit: 1` gives OBJECT when it should stay as ARRAY since index 1 is within limit). That's the failing test you found.

Since the original issue is already resolved and my change introduces an off-by-one regression, I'll close this PR.

PR #559, opened 2026-06-07
[Fix] `utils`: respect encoding of surrogate pairs across chunks

Q1Xuan:
### Summary
- avoid splitting UTF-16 surrogate pairs across the encoder's 1024-code-unit chunks
- add a regression test for a surrogate pair that lands on the chunk boundary
- rebuild the browser bundle

### Bug
`utils.encode` chunks long strings before percent-encoding. When a high surrogate lands at the end of a chunk and the low surrogate is in the next chunk, the pair is encoded as two invalid code points instead of one Unicode character.

Example before this change:

```js
var s = Array(1024).join('a') + '😀';
qs.stringify({ x: s }); // ended with %F0%9F%90%80%F2%90%80%80
```

After this change, the output correctly ends with `%F0%9F%98%80`.

### Tests
- `npx tape "test/**/*.js"`
- `npx tape test/utils.js`
- `npx eclint check lib/utils.js test/utils.js`
- `npx eslint lib/utils.js test/utils.js` (no errors; existing `no-continue` warnings remain)
- verified `dist/qs.js` with the boundary reproduction

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -226,6 +226,13 @@ var encode = function encode(str, defaultEncoder, charset, kind, format) {
     var out = '';
     for (var j = 0; j < string.length; j += limit) {
         var segment = string.length >= limit ? string.slice(j, j + limit) : string;
+        if (j + limit < string.length) {
+            var last = segment.charCodeAt(segment.length - 1);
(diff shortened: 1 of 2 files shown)


ljharb (approved):
reverted the dist change (only releases change that) and added some tests

PR #560, opened 2026-06-09
[Fix] `parse`: throw on unbalanced bracket groups

### LINKED ISSUE #558: 6.15.2 parse behavioral change breaks nested bracket query string parsing
## Summary

Upgrading from `qs@6.15.0` to `qs@6.15.2` introduced a behavioral change in `parse` that breaks existing query string handling in Express-based applications using nested bracket group notation.

## Context

The `6.15.2` changelog mentions:

> `[Fix] parse: handle nested bracket groups and add regression tests (#530)`

While this is framed as a fix, it changes the **parsed output** for query strings that were previously handled differently. Applications relying on the prior behavior see regressions without any code change on their end.

## Impact

In our case, query strings like:

```
groupBy=service&groupBy=linkedaccid&startDate=2021-08-01&...
```

are parsed differently between `6.15.0` and `6.15.2`, causing API endpoints to receive incorrect or empty parameter values, resulting in empty responses where data is expected.

## Suggestion

It would help to:
1. Clearly document this as a **breaking behavioral change** in the changelog (not just a fix), so consumers can assess impact before upgrading.
2. Consider whether the prior behavior was actually incorrect or just undocumented, as many applications may depend on it.

## Environment

- `qs` version: `6.15.0` (working) → `6.15.2` (broken)
- Used via `npm overrides` in an Express.js application
- Node.js version: 24

pupuking723:
Closes #558.

This changes parser behavior for malformed nested keys that start a bracket group after a parent key but never close it. Instead of treating the rest of the key as a literal segment, `qs.parse` now throws a `RangeError`, which matches the maintainer direction in the issue discussion.

Existing behavior is preserved for bracket-prefixed keys and for keys with unterminated inner bracket groups that are already handled as literal segments.

Verification:

- `npx tape test/parse.js`
- `npm run tests-only`
- `npm run lint`
- `npm test`

### DIFF
--- a/lib/parse.js
+++ b/lib/parse.js
@@ -229,6 +229,20 @@ var parseObject = function (chain, val, options, valuesParsed) {
 // qs parse semantics for depth/prototype guards.
 var splitKeyIntoSegments = function splitKeyIntoSegments(originalKey, options) {
     var key = options.allowDots ? originalKey.replace(/\.([^.[]+)/g, '[$1]') : originalKey;
+    var bracketLevel = 0;
+
+    for (var j = 0; j < key.length; ++j) {
+        var code = key.charCodeAt(j);
+        if (code === 0x5B) { // '['
+            bracketLevel += 1;
+        } else if (code === 0x5D && bracketLevel > 0) { // ']'
+            bracketLevel -= 1;
+        }
+    }
+
+    if (bracketLevel > 0) {
+        throw new RangeError('Unbalanced bracket group in query string key');
+    }
 
     // depth <= 0 keeps the whole key as one segment
     if (options.depth <= 0) {
@@ -280,10 +294,7 @@ var splitKeyIntoSegments = function splitKeyIntoSegments(originalKey, options) {
         }
 
         if (close < 0) {
-            // Unterminated group: wrap the raw remainder in one bracket pair so it stays
-            // a single literal segment (e.g. "[[]b" -> "[[]b]"); we do not infer missing ']'.
-            segments[segments.length] = '[' + key.slice(open) + ']';
-            return segments;
+            throw new RangeError('Unbalanced bracket group in query string key');
         }
 
         var seg = key.slice(open, close + 1);
--- a/test/parse.js
+++ b/test/parse.js
@@ -20,6 +20,14 @@ var characterizeParse = function characterizeParse(st, input, opts, expected, la
     st.deepEqual(result, expected, label + ': parses to the current lenient output');
 };
 
+var throwsUnbalancedBracket = function throwsUnbalancedBracket(st, input, opts, label) {
+    st['throws'](
+        function () { qs.parse(input, opts); },
+        new RangeError('Unbalanced bracket group in query string key'),
+        label + ': throws on the unbalanced bracket group'
+    );
+};
+
 test('parse()', function (t) {
     t.test('parses a simple string', function (st) {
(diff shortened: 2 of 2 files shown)


ljharb:
I've added some tests to `main`, and rebased this PR on top of it, to make sure any breaking changes are surfaced.

pupuking723:
Updated the branch after the new characterization tests were added on main.

The PR now keeps the new throwing behavior for unbalanced bracket groups after a parent key, including the #558 reproduction, while preserving the existing lenient behavior for bracket-prefixed keys, inner-bracket cases, and depth-limited literal remainders.

Verification:
- `HOME=/private/tmp/qs-test-home npm run tests-only`
- `npm run lint`

ljharb:
The `\n` in your message strongly implies you're an LLM - can you confirm you're a human and have substantially contributed to the content in this contribution? If not, your contribution is not copyrightable and thus is not viable for OSS.

lib/parse.js
```diff
@@ -280,6 +280,10 @@ var splitKeyIntoSegments = function splitKeyIntoSegments(originalKey, options) {
         }
 
         if (close < 0) {
+            if (parent && key.indexOf('[', open + 1) < 0) {
```
ljharb: If we're going to throw on unbalanced brackets, I think we should throw on *all* of them rather than this narrow subset. As written, the guard only fires when there's a `parent` and no later `[`, which leaves three inconsistent escape hatches:

1. **Depth-dependent.** `a[b]c[d]e[f=v` throws at `depth: 5` but silently falls through to the lenient literal path (below, ~line 312) at `depth: 1`. Whether malformed input throws shouldn't depend on the `depth` option
2. **Later-`[` escape.** `key.indexOf('[', open + 1) < 0` means `a[bc[=v` stays lenient even though it's just as unbalanced - only a *trailing* unclosed bracket throws
3. **Bracket-prefixed keys never throw** (the `parent &&` guard), so `[a[b=v` stays lenient too

So the effective rule is hard to describe: "unbalanced brackets throw, unless the key starts with `[`, or has another `[` after the unclosed one, or the unclosed group is past the depth budget." If the intent is "malformed key → throw," I'd rather detect an unbalanced bracket count for the whole key up front and throw unconditionally, so the behavior is consistent regardless of `parent`, later brackets, or `depth`.

pupuking723:
> 你在消息中的 `\n` 一词显然表明你是一名大型语言模型。你能确认你是人类吗？而且，你确实对这篇内容做出了实质性的贡献吧？如果不是这样，那么你的贡献就不属于版权保护的范围，因此不适合用于开源项目中。

Yes, I am a human contributor and I substantially reviewed and contributed to this change.

I used coding assistance while working on it, but I read the issue, reviewed the implementation and tests, made the final decisions, and take responsibility for the contribution.

pupuking723:
Updated this to validate the whole key for unmatched opening brackets before applying depth splitting. The throw no longer depends on whether there is a parent key, another `[` later in the key, or a lower depth limit.

I also rebased onto the latest `main` and squashed the branch back to a single focused commit.

Verification:
- `HOME=/private/tmp/qs-test-home npm run tests-only`
- `npm run lint`

techouse (approved):
LGTM

pupuking723:
Clarifying my earlier reply: I am the human account owner and contributor. I used AI coding assistance while preparing this change, then reviewed the implementation and tests and take responsibility for what I submitted. I understand if this repository does not accept AI-assisted contributions.

PR #567, opened 2026-07-15
docs: clarify 6.15.2 parse bracket behavior change

### LINKED ISSUE #558: 6.15.2 parse behavioral change breaks nested bracket query string parsing

Hashim1999164:
## Summary

- Note in the 6.15.2 changelog that the nested bracket `parse` fix can change output for some unbalanced bracket keys (#558).
- Add a short README note so upgraders can find the discussion before bumping past 6.15.1.

Fixes #558

## Test plan

- [x] Documentation-only change (no runtime code).
- [ ] Maintainer review for wording and placement.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,7 @@
 - [Fix] `stringify`: use configured `delimiter` after `charsetSentinel` (#555)
 - [Fix] `stringify`: apply `formatter` to encoded key under `strictNullHandling` (#554)
 - [Fix] `stringify`: skip null/undefined filter-array entries instead of crashing in `encoder` (#551)
-- [Fix] `parse`: handle nested bracket groups and add regression tests (#530)
+- [Fix] `parse`: handle nested bracket groups and add regression tests (#530); changes output for some unbalanced bracket keys (see #558)
 - [readme] fix grammar (#550)
 - [Dev Deps] update `@ljharb/eslint-config`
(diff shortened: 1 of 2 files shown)


ljharb (approved):
Combined with #568

PR #568, opened 2026-07-17
docs: clarify 6.14.1 parse bracket behavior change

### LINKED ISSUE #537: Query string array parsing behavior changed in version 6.14.1

hfhchan-plb:
## Summary

- Clarify changelog entry for version 6.14.1 regarding the behavior change of `arrayLimit` and its application to `[]` notation.
- Clarify in README.md that arrayLimit applies to both '[]' and '[0]' syntaxes since v6.14.1.

Fixes #537

## Test plan

- [x] Documentation-only change (no runtime code).
- [ ] Maintainer review for wording and placement.

### DIFF
--- a/README.md
+++ b/README.md
@@ -293,6 +293,8 @@ assert.deepEqual(withIndexedEmptyString, { a: ['b', '', 'c'] });
 Any array members with an index of `20` or greater will instead be converted to an object with the index as the key.
 This is needed to handle cases when someone sent, for example, `a[999999999]` and it will take significant time to iterate over this huge array.
 
+Since v6.14.1, the arrayLimit now applies to ararys specified using both the `[]` and `[0]` syntaxes.
+
 ```javascript
 var withMaxIndex = qs.parse('a[100]=b');
 assert.deepEqual(withMaxIndex, { a: { '100': 'b' } });
--- a/README.md
+++ b/README.md
@@ -293,7 +293,7 @@ assert.deepEqual(withIndexedEmptyString, { a: ['b', '', 'c'] });
 Any array members with an index of `20` or greater will instead be converted to an object with the index as the key.
 This is needed to handle cases when someone sent, for example, `a[999999999]` and it will take significant time to iterate over this huge array.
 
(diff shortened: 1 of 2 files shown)


hfhchan-plb:
If you prefer, I can change the wording "The behaviour of `arrayLimit` has changed and now also applies to `[]` notation." to "The behaviour of `arrayLimit` has been fixed and now applies to `[]` notation." to make your stance clearer on why this change was warranted.

README.md
```diff
@@ -293,6 +293,8 @@ assert.deepEqual(withIndexedEmptyString, { a: ['b', '', 'c'] });
 Any array members with an index of `20` or greater will instead be converted to an object with the index as the key.
 This is needed to handle cases when someone sent, for example, `a[999999999]` and it will take significant time to iterate over this huge array.
 
+Since v6.14.1, the arrayLimit now applies to ararys specified using both the `[]` and `[0]` syntaxes.
```
hfhchan-plb:
```suggestion
Since v6.14.1, the arrayLimit now applies to arrays specified using both the `[]` and `[0]` syntaxes.
```

ljharb (approved):
Combined with #567

PR #582, opened 2026-07-27
chore: exclude test files from publishing

gameroman:
I think `test/` files shouldn't be published to npm, it looks like they were included accidentally

Let me know if I'm wrong or there's a reason to publish

I'd also exclude `CHANGELOG.md`, but that's likely a separate discussion


ljharb (changes requested):
Like all my packages, they're included intentionally, so you think wrong. `npm explore foo && npm install && npm test` should always work.

ljharb:
Duplicate of #400 and many others.

PR #592, opened 2026-08-10
Optimize flat query parsing

sarthakagrawal927:
## What changed

Flat query strings now bypass per-key `parseKeys` / `parseObject` construction and `utils.merge` calls. The parser first proves that every parsed key can preserve its existing top-level meaning, then copies the already-parsed values into the result.

Nested keys, dotted notation, decoded-dot keys, object input, empty keys, prototype-sensitive keys, and values that still require array compaction retain the existing path.

## Why

The normal parser path builds and merges a one-key object for every parameter, even when a query contains only independent top-level keys. That repeated object construction and merging dominates large flat query strings.

## Performance

CodeVetter ran ten paired, interleaved processes per revision after one warmup, using the identical Node test and ten parses per input size:

| Parameters | Baseline | Candidate | Change |
| ---: | ---: | ---: | ---: |
| 1,000 | 0.772 ms/op | 0.518 ms/op | -32.902% |
| 10,000 | 9.266 ms/op | 4.880 ms/op | -47.334% |
| 40,000 | 41.706 ms/op | 23.183 ms/op | -44.413% |

The endpoint scaling exponent moved from 1.081 to 1.030. The paired verifier returned `confirmed`, `mechanically_confirmed: true`, `materially_useful: true`, and `shipping_recommended: true`, with no recorded limitations.

These numbers were measured on Node 24.19.0 / Apple M5 Pro and are intended to show the relative implementation effect, not hardware-independent throughput.

## Correctness and safety

Broader direct-copy variants were rejected during development because the upstream suite caught mixed array/scalar, object-input, and decoded-dot semantic changes. The retained preflight falls back before copying any value when a key requires structural parsing.

- all 1,045 upstream tests pass
- 100% statement, function, and line coverage
- ESLint and EditorConfig checks pass with the repository's existing warnings only
- README evaluation passes
- distribution build passes
- `git diff --check` passes

The proposed commit changes only `lib/parse.js`; the benchmark harness is not included.


ljharb:
@sarthakagrawal927 by deleting the fork, this PR is now unrecoverable. Please don't delete forks with open PRs in the future.

sarthakagrawal927:
yes, thanks for pointing out. My agent must have deleted this fork accidentally. Please let me know, if you think the code is helpful. I will re-fork & reopen.

ljharb:
it seems helpful, but i haven't had a chance to review it yet.

PR #593, opened 2026-08-13
[Fix] stringify: serialize Date values returned unchanged by a filter function

### LINKED ISSUE #467: Serialization for `Date` is not working when using `filter` option.
serialization for `Date` type values are skipped when I'm usinng `filter` option for custom serialization.

### code to reproduce

```ts
import qs from 'qs'

class Range {
  constructor(
    public from: number,
    public to: number
  ) {}
}

const data = {
  a: new Range(1, 123),
  b: new Date(),
  c: [new Date(), new Range(3, 45)],
  d: false,
  e: { a: 123, b: new Date(), c: new Range(6, 700) },
}

const stringified = qs.stringify(data, {
  filter(_, value) {
    if (value instanceof Range) {
      return `${value.from}...${value.to}`;
    }
    return value;
  }
});

console.log(qs.parse(stringified));
```

### what I expected

```js
{
  a: '1...123',
  b: '2023-03-07T09:02:41.260Z',
  c: [ '2023-03-07T09:02:41.260Z', '3...45' ],
  d: 'false',
  e: {
    a: '123',
    b: '2023-03-07T09:02:41.260Z',
    c: '6...700'
  }
}
```

### what actually worked

```js
{
  a: '1...123',
  c: [ '3...45' ],
  d: 'false',
  e: { a: '123', c: '6...700' }
}
```

so I have to call `qs.stringify` like this:

```ts
const stringified = qs.stringify(data, {
  filter(_, value) {
    if (value instanceof Range) {
      return `${value.from}...${value.to}`;
    }

    // I have to serialize `Date` by my self
    if (value instanceof Date) {
      return value.toISOString()
    }

    return value;
  }
});
```

deepshekhardas:
When a `filter` function returns a `Date` value unchanged, it was previously serialized as the empty string (its own enumerable keys) and silently dropped from the output. Now `serializeDate` is applied to the filtered value, matching the no-filter path.

```js
const date = new Date();
qs.stringify({ a: date }, { filter: (prefix, value) => value });
// was: ''
// now: 'a=...'
```

Closes #467


ljharb:
Looks like this is a duplicate of #595.

PR #594, opened 2026-08-13
[Fix] stringify: serialize Set values as arrays

### LINKED ISSUE #264: Add support for sets
If the object passed to the `stringify` method holds a [Set](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set), the whole key and value are ignored by the stringification process. A custom encoder can't be used either, because it doesn't receive the set-typed entry.

deepshekhardas:
`qs.stringify` silently dropped `Set` values — the key/value pair was omitted entirely because a `Set` has no enumerable own keys.

This PR normalizes `Set` instances to arrays during stringification, both at the top level and when nested, so they serialize using the same array format options as plain arrays.

```js
qs.stringify({ a: new Set([1, 2, 3]) });
// 'a%5B0%5D=1&a%5B1%5D=2&a%5B2%5D=3'
```

Closes #264


lib/stringify.js
```diff
@@ -87,6 +87,10 @@ var stringify = function stringify(
 ) {
     var obj = object;
 
+    if (obj instanceof Set) {
```
ljharb, line 90: this will break on engines that lack `Set`

ljharb (changes requested):
We can't rely on the existence of Set, or Array.from, or the iterator protocol. If anything, we'd have to conditionally cache Set.prototype.forEach at module level, and use that (the iterator protocol should be avoided).

